PR #1, opened 2020-12-02
-Added Spellbinder Driver

jessaskey:
I will email Carsten the ROM images that should match the CRM and SHA1 hashes in the driver here. These are BETA ROM's.


freezy:
Looks like this made it in, thanks! 👍

PR #4, opened 2021-02-04
libpinmame: rename dll folder to libpinmame. Add GetGames API function

jsm174:
This PR is some general cleanup for libpinmame.

- Renames the `dll` folder to `libpinmame`. 
- Moves `ios` ticker to `libpinmame` and is shared between linux and osx. (Win still uses `win/ticker.c`)
- Added `GetGames()` API function that is used to retrieve a list of support games
- Added `cmake/libpinmame` folder and added github workflow actions to automatically build libpinmame.


toxieainc:
Neat! Thanks!

PR #12, opened 2021-02-19
ci: add a prerelease workflow. move installer into vpinmame.

jsm174:
This PR adds a new workflow, `prerelease`. Whenever the code is ready for testing and has completed building, this can be launched manually by going to `actions` -> `prerelease` -> `Run workflow`.

We added an optional SHA field. If we want to make a pre-release from a commit that is not HEAD, that commit SHA can be entered here.

Screen Shot 2021-02-19 at 8 46 56 AM

Two other small changes included in this PR.

1) Do not pack the installers with UPX. The installer ends up small, and windows may flag it as a virus.
2) The instvpm workflow is now migrated into the vpinball workflow. This allows the installer and setup to be included in same artifact.


toxieainc:
🥇

PR #65, opened 2022-08-16
Unixsdl

gnulnulf:
I added makefiles for unix-sdl and unix-sdl with debugging support.
I also created a little readme how to use them
I tested this on ubuntu 18.04 and 20.04.


toxieainc:
Great, thanks!

PR #69, opened 2022-08-30
fixed some typos in LISY readme files

mkalkbrenner:


toxieainc:
Thanks! Although i don't know if the 'HOme' spelling variant is actually intentional for whatever reason.

mkalkbrenner:
@bontango I think the official name is LISY_Home, right?

toxieainc:
Also true.. ..

bontango:
thx for corrections, and yes it is 'LISY_Home' or 'LISY_H'
btw. in the readme you mentioned Atari support with LISY. 
This is (still) planed, but not implemented yet.

toxieainc:
@bontango : All the lisy files are more or less a 1:1 copy of your fork, see https://github.com/bontango/lisy-src-5.x/blob/master/src/lisy/README.lisy#L1 , so the Atari part is coming from you directly.  ;)
I just always tried to keep everything up to date in 'our' copy, as obviously PinMAME still continues to have changes in GTS/S11/etc, but your fork is based on a fixed version.

PR #71, opened 2022-09-15
Lisy update

mkalkbrenner:


mkalkbrenner:
The workflow results could be seen at https://github.com/mkalkbrenner/pinmame/actions
especially https://github.com/mkalkbrenner/pinmame/actions/runs/3062043588

jsm174:
Awesome stuff! 

If you'd like these artifacts to be included when the prerelease workflow gets kicked off, add it here:

https://github.com/vpinball/pinmame/blob/b2ae4a9b11ab03cc64ba3e724454792a04ca56f4/.github/workflows/prerelease.yml#L42

mkalkbrenner:
Just a short explanation:
- We re-formatted the code of the lisy subfolder to be able to to compare different forks and to merge them.
- The current version within PinMAME missed some changes and could not be built. That's fixed.
- I added a workflow to ensure that lisy could still be if other things in PinMAME change.
- After the PR has been merged, both projects should be kept in sync using git mechanisms instead of copy&paste like in the past.

> If you'd like these artifacts to be included when the prerelease workflow gets kicked off, add it here:
Good to know! In that first PR I focused on the "testing" part to ensure that a push to the PinMAME repository dosn't break things. LISY itself requires more then the special PinMAME version.
We're already working on an automated build of Raspberry Pi images. But this should happen in a dedicated project.
So we can still decide whether the workflow here should be included in a release process or not.

bontango:
Many thanks to mkalkbrenner for the work done. I was in contact with him and I'm fine with the corrections made to my LISY code.

toxieainc:
One more thing before waiving this through:
Would it be okay to have all this stuff under the BSD3 license?
Cause with PinMAME our goal is to ideally match MAME, which is going the BSD3 route.
@mkalkbrenner @bontango

mkalkbrenner:
The code has been re-formatted and received some bug fixes. So nothing new and its old license still applies.
From my point of view it was and still is BSD-3.

mkalkbrenner:
Only the workflow to build lisy is new and I agree to contribute it under the main license of the project.

bontango:
I'm fine with BSD3. Do I need to add anything to the header of my files?

mkalkbrenner:
Since we didn't change the amount of files under src/lisy/ I suggest to merge this PR as it is.
If required we can add license headers later. But in general the license should be covered by https://github.com/vpinball/pinmame/blob/master/LICENSE for all subdirectories as long as there's not mentioned something else within a file or directory.

PR #85, opened 2023-01-05
Fix WPC GI Dimming

vbousquet:
When working on Twilight Zone and Creature From The Black Lagoon, I had problem with getting correct GI dimming. I noted that other existing VPX versions of these tables (and lots of WPC others) have some hacks around the GI dimming.

After some investigation, it appears that there were 3 problems with the existing implementation:
- it would use the data written to the ASIC register, filtering some of it, this hack would work as long as only one GI line would be faded but fails when there are more than one.
- it was not able to evaluate between 7 and 8 out of the 8 PWM levels (full cycle or first IRQ write)
- it was very slightly incorrect regarding fading since it takes for granted that there are 8 IRQ per Zero Cross cycle, which is wrong: it's most of the time 8 but sometimes 9 (no noticeable effect, just to be exact).

After searching some more (mainly stepping through CFTBL and also reading FreeWPC implementation), I think that the implemented behavior was wrong. My understanding of the ASIC behavior is the following (and implemented in this PR):
- The ASIC has a Zero Cross latch which is turned on when the zero cross signal is received (120Hz in US, so a bit more than 8ms)
- When the ASIC receive a read command from the CPU on the WATCHDOG register and the Zero Cross latch is high, it will reset the GI triac outputs to a stored value (and also reset the Zero Cross latch)
- When the ASIC receive a write command from the CPU on the GI LAMP register, it will store the data for the next zero cross reset, but it will also raise up any output triac line for which the bit is set

This behavior is used by the CPU for dimming by  having a GI reset value at 0 (Read WATCHDOG with ZC), then rise up the GI line after a few IRQ (Write GI LAMP at 1), then prepare the ASIC for the next cycle by immediatly writing the GI line at 0 (Write GI LAMP at 0) which is not outputed until the next zero cross watchdog read.

The existing hack would skip the second write, which works when there is only 2 write in a zero cross cycle but would fail if there is more (for example 2 GI lines fading).

I made the PR to be backward compatible (outputs 0 to 8 level). I have tested it on CFTBL, TZ, T2 and MM. All looks fine but it is difficult to be absolutely sure.

### DIFF
--- a/src/wpc/wpc.c
+++ b/src/wpc/wpc.c
@@ -16,6 +16,7 @@
 
 #define PRINT_GI_DATA      0 /* printf the GI Data for debugging purposes   */
 #define DEBUG_GI           0 /* debug GI code - more printf stuff basically */
+#define DEBUG_GI_W         0 /* debug GI write - even more printf stuff */
 #define WPC_FAST_FLIP      1
 #define WPC_VBLANKDIV      32/* This steers how precise the DMD FIRQ interrupt is (as it depends on the DMD_FIRQLINE) */
                              /* Best should be 64, but this leads to instability for some machines DMD (e.g. T2)      */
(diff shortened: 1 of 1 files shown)


toxieainc:
Both gaston and me are no experts of the details there, so we just trust you that all is good  ;)

PR #88, opened 2023-01-12
Better WPC GI dimming

vbousquet:
I had a doubt regarding the result of my previous commit for WPC GI dimming in a few situations, so I added a high frequency inspector (will propose a PR later when it will be cleaned out) and get the following results:
image

(Solenoids are the top lines, GI the 5 bottom ones, sampling frequency is IRQ so around 1ms). It shows that GI is mostly ok but has some 'holes' (the 3 blue ones in the bottom line). So I checked back, @mkalkbrenner gave me this very good reference: https://github.com/bitfieldlabs/aggi/tree/master/scopes which allowed me to discover one error of my implementation. I was under the assumption the ASIC used latching logic and thyristor but in fact these are just outputs wired to Triac (which makes perfectly sense with the zero cross logic). So this PR fixes the code and is true to the schematics.

The sad bit is that even if it is better (and therefore, I think should be merged), it won't solve the little holes that are sometimes seen. These are very seldom and if using a light PWM integrator difficult to see but it is still a little timing bug.

### DIFF
--- a/src/wpc/wpc.c
+++ b/src/wpc/wpc.c
@@ -15,7 +15,7 @@
 #endif
 
 #define PRINT_GI_DATA      0 /* printf the GI Data for debugging purposes   */
-#define DEBUG_GI           0 /* debug GI code - more printf stuff basically */
+#define DEBUG_GI           0 /* debug GI PWM code - more printf stuff basically */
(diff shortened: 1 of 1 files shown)


toxieainc:
Again, great detective work! Thanks!

Issue #66, opened 2022-08-19
Modulated solenoids in SE hardware


apophis79:
Currently pinmame does not support modulated solenoid outputs in Stern/SEGA Whitestar (SE) machines, despite those machines having this capability. It would be great to be able to use modulated signals with these machines.

vbousquet:
PR #91 adds PWM integration which is a similar feature to modulated solenoids, giving user more control on the way the modulation is computed. So SE hardware do have modulated outputs now.

@toxieainc I think this can be closed for now. @apophis79 please reopen it if the new feature does not fully cover this request.

toxieainc:
Thanks!

PR #91, opened 2023-01-22
Improve new PWM integration

vbousquet:
This PR:
- fix when PWM integration is used with a system that does not have modulated solenoids implemented,
- add support for WhiteStar and SAM
- performs a few cleanups

S11 is tested and fully working (with a full table configured with these new PWM outputs). Whitestar was tested with the graphic analyzer and seems good. SAM was also tested with the graphic analyzer and looks good but, perhaps, more testing for SAM (with a full table) could be interesting since the integration frequency is fairly different (4kHz instead of 1kHz for other systems)

### DIFF
--- a/src/wpc/core.c
+++ b/src/wpc/core.c
@@ -2324,7 +2324,7 @@ void core_perform_output_pwm_integration(core_tModulatedOutput* output, int nSam
 void core_perform_pwm_integration()
 {
    int nSamples = coreGlobals.pulsedOutStateSamplePos - coreGlobals.lastModulatedOutputIntegrationPos;
-   if (coreGlobals.nModulatedOutputs > 0 && coreGlobals.pulsedOutStateSampleFreq > 0 && nSamples > 0)
+   if (coreGlobals.pulsedOutStateSampleFreq > 0 && nSamples > 0)
    {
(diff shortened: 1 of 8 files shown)


toxieainc:
Thanks once more.
btw: I don't know how much this is related to all the work you're doing right now, but there are (at least) four more things that could well be:
a) https://github.com/vpinball/pinmame/issues/66
b) Capcom flipper solenoid strength (e.g. https://www.vpforums.org/index.php?showtopic=33669&p=331731 and following)
c) Early WPC (e.g. BOP) display flicker/dimming/'fade'
d) same for GTS3

vbousquet:
These are likely directly connected to this feature. The point is the way (or the lack) PinMame emulates the non digital part of the hardware (coils, bulbs,...):
a) Implemented in this PR (at least a similarfeature with the same intent), so should be good to close
b) I would guess this is exactly this. I have identified quite a lot of this type of behaviors for WPC as well. When solenoids do not have a hold coil, the hardware will modulate with PWM to avoid burning the coil, or to dim the strength. To go further, this would need to add solenoid integrators based on solenoids characteristics, and output a value with regards to the strength ratio (no difficulty, but needs a bit of work and time)
c) this has to be looked more precisely, may be the need of an integration or timing issues in the emulation
d) same as c

PR #94, opened 2023-01-26
Improve Incandescent Bulb model

vbousquet:
While implementing the companion feature for VPX, I identified a few precision issue that lead me to refactor the incandescent model for a cleaner version, namely:
- fix precision issues (due to emission polynomial regressions being 3rd order),
- improve performance by precomputing most of the maths
- cleanup implementation so that it can be shared accross PinMame / VPX / VPE / whoever wants it :)
- adjust bulbs characteristics so that they all stabilize at 2700K
- add an implementation of the model for efficient steady state (mainly for VPX)

I don't know what is the right license for that (since the bulb.c/bulb.h files are meant to be shared across projects), so I put the same comment that what is done in VPX for the part that is shared with VPE. Feel free to change to whatever suits PinMame licensing.


toxieainc:
Nice, thanks!

PR #109, opened 2023-04-17
missed openAL closeDeivce call

shprotru:
Greetings, I am new! I tried to run the example from the repository and got a message about a problem related to OpenAL:
```
build/Release/ppuc -c src/ppuc/examples/t2.yml
AL lib: (EE) alc_cleanup: 1 device not closed
```
Correct roms I put on their path. There was no other error messages, just silent launch. My operating system is Linux Ubuntu 22.04.2 LTS. Then I casually familiarized myself with the code and found that call to `alcCloseDevice` is missed. Also I added 2 additional checks for `alcOpenDevice` and `alcCreateContext` calls which shuld help to diagnose the problems related to OpenAL.


toxieainc:
Great, thanks!

PR #114, opened 2023-05-07
Roscoe build fixes 20230503

droscoe:
I've made changes to the CMake files for all Windows projects to support using the workflow scripts to generate user-level project files that successfully compile in both RELEASE and DEBUG mode.  Ultimately, this will allow project files to always stay in step with the build, and enable the removal of the project files in source control.  For now, however, it is a first step to allow DEBUG builds with minimal changes.

I had to add a non-functioning definition for logerror in the config.c file, since it is not being provided by the other projects, such as VPinMAME and was causing linking errors.  A future pull request will make this logging function work in the standalone lib.

I have tested the CMake files and confirm they build without error on my machine for all Windows projects in RELEASE and DEBUG mode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .project
+build/
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .project
 build/
+/CMakeLists.txt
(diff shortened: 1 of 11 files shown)


toxieainc:
Looks good, thanks!!

droscoe:
> Looks good, thanks!!

Damn!  I see the build is failing.. I will fix it right away..

PR #117, opened 2023-05-12
Windows universal build script

droscoe:
These changes include:
- Small tweak to src/windows/config.c to fix DEBUG build
- Windows batch script to automate developer builds

The script uses the same CMake scripts that the workflows use, to create Windows project files for all of the PinMAME projects.  I have tested every configuration of the build through the script (Debug, Release, Win32, x64, SAMCOLOR), and everything builds without any additional warnings or errors.  I have included comprehensive help to explain the script options.  It is accessible directly by opening the "win_build_help.txt" file as well as in-script via the "/?" option.

In theory this COULD replace the version-controlled project files that are in the repository.  Users will need to have CMake installed in addition to the normal dependencies for Visual Studio.  It has a limitation in that it can only create the individual project files for each application.  It cannot yet create a single solution to build all of them.  That will be added at some point, if possible.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -650,13 +650,16 @@ int config_write_mixer_config(config_file *cfg, const struct mixer_config *mixer
 	return CONFIG_ERROR_SUCCESS;
 }
 // DAR_DEBUG @20230506 Defining this for now without a filename just to avoid
-//                     linking errors for DEBUG builds
+//                     linking errors for DEBUG builds.
+// DAR @20230510 Added check for LIBPINMAME since we only want this if we are
+//               building libpinmame standalone in DEBUG
 //
 // logerror
 //
 // VPM defines its own logger, only use this in standalone build
 #ifndef VPINMAME 
 #ifndef __GNUC__
+#ifdef LIBPINMAME
 	#if (!defined(PINMAME) || defined(MAME_DEBUG) || defined(_DEBUG)) // In PinMAME, log only in debug mode.
 //		FILE *config_get_logfile(void) { return errorlog ? logfile : NULL; }
 		static FILE *logfile = NULL;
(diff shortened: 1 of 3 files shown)


toxieainc:
As i do not have much clue about all of this, @jsm174 please have a look

jsm174:
If I'm understanding this correctly, technically all the vcxproj, vcxproj.filters, sln, and dsp files could be removed from the repo, and then regenerated via the batch script?

If so, pretty cool. FWIW, I don't really work on the Windows side, so I would defer to others who actively use the project files as your workflow would be impacted. @toxieainc @volkenborn

droscoe:
> If I'm understanding this correctly, technically all the vcxproj, vcxproj.filters, sln, and dsp files could be removed from the repo, and then regenerated via the batch script?
> 
> If so, pretty cool. FWIW, I don't really work on the Windows side, so I would defer to others who actively use the project files as your workflow would be impacted. @toxieainc @volkenborn

TL;DR  My batch file uses the same CMake scripts as the automated CI process.  As long as those work, the generated projects should also work.   In that regard, the version-controlled counterparts are no longer necessary, and technically could be removed. 

Further reading...

The reason I created this script is because I could not get a working build using the existing project files.  I examined the workflows and realized that CMake is already in use there, and those builds work.   With just a little bit of effort, I was able to use those files and generate project files that work, and will always work., as long as the automated CI builds do. 

Assuming that any impacted devs already have a working build, the only new requirement is installing CMake.   The output paths will be slightly different than before, so if they have any post-build automation, it would need to be updated.

**What I propose**
- Accept my changes.  It won't break anything already there.

Optional:
- Move the existing files to a folder called "deprecated_projects" off of the root
- add a README.txt file encouraging Windows devs to install CMake and use the new script
- If they still want to use the old project files, they can copy to root and proceed as usual
- recommend adding .gitignore rules to prevent check-ins of project files

If anyone is having an issue with the new script, I am happy to provide support.

droscoe:
@toxieainc  I spoke with @jsm174 last night, and he agrees that my script has no impact on existing repo other than adding a new capability, and is OK with merging it.

toxieainc:
Thanks to the 2 of you!

PR #124, opened 2023-05-25
HC55516 - add original chip logic

mjrgh:
This adds the original logic implemented in the HC55516 hardware, as determined by the 2020 decap analysis, by way of the MAME update based on the decap.  This should now be an accurate re-creation of the original chip's behavior, with exactly the same outputs as far as the on-chip DAC stage.

I've actually had this code working for a while in a side project, so I'm pretty confident in its stability as far as the new decoding algorithm goes.  I've only lightly tested this particular integration, though, just enough to make sure it's working.

The original code is still included and can be selected (it's even run-time configurable, although that's not exposed at a user level anywhere).

### DIFF
--- a/src/sound/hc55516.c
+++ b/src/sound/hc55516.c
@@ -27,6 +27,40 @@
 // as of this writing, although the chip has been out of production for a
 // very long time.
 //
+// Syllabic filters are defined in analog terms, but like any analog
+// (that is, continuous-time) filter, they can be translated into the
+// digital (discrete-time) domain.  The HC555xx chip was in fact a
+// discrete-time implementation of the CVSD filter system.  A recent (2020)
+// decap analysis of the HC55516 provided us with the exact algorithm and
+// arithmetic parameters, which we now implement here.  The old MAME
+// implementation was also a discrete-time implementation, implementing
+// a digital version of the analog filter system specified in the data
+// sheet.  The MAME version used IEEE floating-point types to implement
+// the filter using the standard discrete-time algorithm for this type
+// of filter.  Based on the decap, we can see that the HC555xx used the
+// same mathematical model, but used 10- and 12-bit fixed-point arithmetic
+// registers to perform the same calculations.  The arithmetic works out
+// to be the same after taking into consideration the lower precision of
+// the original hardware's registers, which shouldn't be surprising given
+// that it would have to be essentially identical to work at all.  The
+// old MAME version is arguably the "better" of the two implementations,
+// in that it simulates the reference filter with greater mathematical
+// precision, thanks to the higher precision of its intermediate results
+// and accumulators.  The extremely limited precision of the original
+// hardware results in greater rounding errors.  However, replicating
+// the bit-for-bit identical arithmetic produces a more exact replica
+// of the original output.  The difference in precision is probably too
+// small to be audible even in theory, and certainly doesn't seem to be
+// in practice.  The one significant difference between the two
+// implementations is that we know from the decap what the clipping
+// behavior is at the limits of the accumulator register ranges, which
+// the old MAME implementation wasn't able to replicate, as it's not
+// something we could have inferred from the reference filter spec in
+// the data sheet.  The lack of proper clipping in the old PinMame
+// implementation motivated the addition of a loudness compression
+// adjustment to handle the excessive dynamic range that resulted.  That
+// might actually have been detrimentally audible.
+//
 // The HC55516 was part of a family of nearly identical chips that also 
 // included the HC55536 and HC55564.  Apparently the only difference among
 // these variations was the maximum sample clock rate supported, so they're 
@@ -110,6 +144,53 @@
 
 // ---------------------------------------------------------------------------
 //
+// Implementation selection
+//
+// There are now two implementations of the HC555xx family:
+//
+// * The original PinMame implementation, which uses the standard discrete-
+//   time filter algorithms to create a model of the ideal syllabic filter
+//   per the published specifications in the HC55516 data sheet
+//
+// * A new HC555xx logic simulation, which re-creates the exact bit logic
+//   implemented in the HC555xx chips, as determined by a 2020 decap
+//   analysis of a specimen of the physical chip.
+//
+// The two implementations perform essentially identical math, since they're
+// both implementing the same discrete-time filter system.  The old PinMame
+// version implements the math using native 'double' types, whereas the chip
+// logic uses 10- and 12-bit 2's complement integer registers (which aren't
+// actually integer values mathematically throughout the calculation: the
+// chip is also doing fractional arithmetic, but doing so using explicit
+// bit shifts rather than implicitly using C operators with floating-point
+// types).  The old PinMame version executes the intermediate calculations
+// at higher precision than the chip, due to the wider registers it uses,
+// but at the cost of some discrepancy from the exact bit values produced
+// by the chip, due to the greater rounding error with the chip's narrow
+// registers.
+//
+// You can select which implementation you want here.  The selection can
+// be changed at run-time, although at the moment this isn't exposed as a
+// user-configuration parameter.  The selection must be made prior to
+// initialization, since it's fixed into the chip state structures once
+// those are created.
+//
+// The two versions sound identical subjectively.  The old PinMame version
+// is a more accurate model of the abstract mathematical filter that the
+// chip purports to implement, but the hardware logic version is a more
+// accurate re-creation of the chip, and should exactly match its output
+// as far as the chip's internal DAC stage.  (We obviously can't truly
+// replicate anything beyond that, since the rest of the hardware system
+// is in the analog domain.)
+//
+//   0 -> use the original PinMame implementation
+//   1 -> use the chip logic simulation
+//
+int hc55516_use_chip_logic = 1;
+
+
+// ---------------------------------------------------------------------------
+//
 // ***** CLOCK RATE MEASUREMENT AND LOGGING *****
 //
 // The HC55516 doesn't have a fixed sample data rate; the rate is simply
@@ -176,7 +257,7 @@ static FILE *sample_log_fp;
 // (<ROM NAME>.vpm_hc55516_dynrange.log) containing periodic updates with
 // the maximum sample level seen so far, and the computed "max non-clipping 
 // gain" (the gain that maps the maximum sample level to PCM full scale).
-// 
+//
 //#define LOG_DYN_RANGE
 
 #ifdef LOG_DYN_RANGE
@@ -187,6 +268,7 @@ static FILE *dynrange_log_fp;
 #endif // LOG_DYN_RANGE
 
 // ---------------------------------------------------------------------------
+//
 // Syllabic filter constants.  For an RC filter, the charge factor per
 // clock period is given by:
 //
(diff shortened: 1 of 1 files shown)


mjrgh:
Note to anyone who wants to test it: you'll have to temporarily switch back to the old YM2151 code, because the new YM2151 code apparently breaks the HC55516 implementation (all versions - independently of this chip logic update).  You have to delete HAS_YM2151_YMFM from the compiler command-line defines, and change the following #defines in pinmame.h:

#define HAS_YM2151_YMFM 0
#define HAS_YM2151_ALT 1

toxieainc:
Yes, this is really weird. I didn't even notice so far that speech is a bit 'slower' than before.
   
  
    
   
  
   Any chance you could have a look at how i wired the new YMFM core up?
   
  
   The majority is within ymfm_opm_pinmame_specific.h
   
  
   In MAME its a bit more sophisticated, so maybe i missed an important detail with the IRQs or how the sound updates are triggered?
   
  
    
   
  
   And i like the way you ported the HC55516 decap core, this is so much more readable than what is in MAME.
   
  
    
   
  
   Thanks!
   
   
   
    mjrgh ***@***.***> hat am 25.05.2023 03:07 CEST geschrieben:
    
   
     
    
   
     
    
     
   Note to anyone who wants to test it: you'll have to temporarily switch back to the old YM2151 code, because the new YM2151 code apparently breaks the HC55516 implementation (all versions - independently of this chip logic update). You have to delete HAS_YM2151_YMFM from the compiler command-line defines, and change the following #defines in pinmame.h: 
   #define HAS_YM2151_YMFM 0#define HAS_YM2151_ALT 1

mjrgh:
> Yes, this is really weird. I didn't even notice so far that speech is a bit 'slower' than before.
> Any chance you could have a look at how i wired the new YMFM core up?

It looks like it's the exact same issue as the old PREDCS_FIRQ_HACK again, if you remember that.  It's the emulation inaccuracy in the original YMFM code where it cleared the Clock A (FIRQ) signal with zero latency after a status register update, whereas the true hardware apparently had a latency significantly greater than 2us (one M6809 cycle).  I say "apparently" because the latency is undocumented, but we can deduce its existence (and a lower bound for the timing) from the way the Williams M6809 sound board software is written.  (I made a diligent effort when we first discovered this to find documentation of the latency in the data sheet and application manual, but I found none. It's just an empirical fact of the hardware implementation, not a documented feature.  But Williams noticed it, either accidentally or on purpose, and took advantage of it ruthlessly, leaving us stuck with replicating it if we want this stuff to work properly.)  The solution in the old YMFM emulator was to schedule the Clock A output clear on a 3us delay, using a MAME timer, instead of executing it in-line with the register update.  The new code naturally doesn't implement the latency, because the latency isn't an official "feature" and no one outside of PinMame has ever noticed it or would ever care.

I haven't quite figured out how the new YMFM class wends its way to the register clear.  If you're familiar enough with that code to isolate it, you'd just have to insert a MAME timer setup where the Clock A bit gets cleared, and send the FIRQ CLEAR in the timer rather than in-line in the same routine.  If you look at the old ym2151.c and search it for IRQ_clear_timer, you'll see the same steps in that code.

Regards,
Mike

________________________________

mjrgh:
I've come up with a possible solution that doesn't require changing any of the YM2151 code.  It would be more "correct" to do it in the YM2151 code, because the latency is pretty clearly an authentic property of the original YM2151 hardware, which as far as I'm concerned means that YM2151 emulators that don't implement the latency properly aren't accurate re-creations.  But the practical reality is that no one except PinMame is ever going to notice or care, so the third-party YM2151 emulators are never going to get it right by themselves.  And given that PinMame now seems to have three separate YM2151 emulators, apparently it's a weirdly popular pass-time to write new ones, so I guess this is going to keep coming up time and again if we expect the YM2151 code to handle it.  So either we keep fixing the new YM2151 emulators every time there's a new-and-improved one to upgrade to, or we fix it somewhere less "third-party".

Here's a solution that places it in what I consider the second most logical place: in the WPC89 sound board code.  That's purely PinMame code, so it's probably less likely that someone will get tempted to upgrade it to some new third-party code that's not sweating all the little details.  And it's very well isolated to the particular combination of emulated chips, so it won't affect anyone except WPC89 games.  It's also very simple and easy to understand.  The idea is that we insert a new WRITE_HANDLER for the YM2151 register writes for the DCS89 sound board only, and in this handler, we monitor for the particular byte sequence that triggers this bug.  When we see it, we execute the problematic register update on a 3us delay timer.  It's a little bit of a hack, but just a little bit.  It would be cleaner to do it in the YM2151 code because that code has a full view of the whole chip state, so it can cancel the timer in oddball cases where the register write is immediately countermanded.  But I don't think that will ever come up in practice with the DCS89 code, so I think this approach will be 100% solid in fact.

The modified code is below (wmssnd.c, starting at line 997).  In light testing it fixes it nicely in my build.  (I might have gone a little overboard on the editorializing in the comments.)

Regards,
Mike


static void DCS89_YM2151_defer_register_14_write(int param)
{
      YM2151_word_0_w(0, 0x14);
      YM2151_word_0_w(1, param);
}

static WRITE_HANDLER( DCS89_YM2151_word_0_w )
{
      // even offset -> register address write; odd offset -> register data write
      if ((offset & 1) == 0)
      {
            // register address write - the data value is the register number, which
            // specifies which register the next odd-offset write affects
            locals.ym_reg_addr = data;
      }
      else
      {
            // Register data write.  If we're writing value 0x15 to register 0x14,
            // defer the write for 3us to simulate the latency of the original
            // YM2151 hardware.  The WPC89 ROM software contains a hard-coded
            // dependency on this latency being longer than the M6809 instruction
            // cycle time.  The M6809 program executes this write while FIRQ is
            // disabled in the status register, then immediately executes an
            // ANDCC instruction that re-enables the FIRQ.  The code sequence
            // assumes that an FIRQ that was asserted while the interrupts were
            // masked out will still be asserted one instruction cycle later,
            // after the ANDCC finishes.  If that assumption doesn't hold, the
            // sound board emulation won't work properly - we get the timing
            // error described in the documentation above regarding the
            // PREDCS_FIRQ_HACK.  The *correct* solution is to change the
            // YM2151 emulator code so that it faithfully replicates the
            // original hardware behavior.  That was the original approach, but
            // it has proved problematic because PinMame now has three different
            // YM2151 emulator implementations (apparently people can't stop
            // writing new ones), and this register-write latency effect doesn't
            // seem to be important to anyone except PinMame, so none of those
            // emulator re-writers bother to properly implement this.  It's not
            // a documented feature, and you wouldn't notice that it even exists
            // without either a very careful analysis of a physical specimen of
            // the chip OR a reference system that demonstrates the behavior,
            // such as a WPC89 board.  At any rate, since the YM2151 emulators
            // can't seem to get it right, a next-best approach is to handle it
            // here, in the bridge bertween the WPC89 emulation and the YM2151
            // emulation.
            if (locals.ym_reg_addr == 0x14 && data == 0x15)
            {
                  // set a one-shot timer to do the deferred register write
                  timer_set(3.0e-6, data, DCS89_YM2151_defer_register_14_write);
                  return;
            }
      }

      // pass it along to the YM2151 emulator
      YM2151_word_0_w(offset, data);
}

static MEMORY_READ_START(wpcs_readmem)
  { 0x0000, 0x1fff, MRA_RAM },
  { 0x2401, 0x2401, YM2151_status_port_0_r }, /* 2401-27ff odd */
  { 0x3000, 0x3000, wpcs_latch_r }, /* 3000-33ff */
  { 0x4000, 0xbfff, CAT2(MRA_BANK, WPCS_BANK0) }, //32K
  { 0xc000, 0xffff, MRA_ROM }, /* same as page 7f */  //16K
MEMORY_END

static MEMORY_WRITE_START(wpcs_writemem)
  { 0x0000, 0x1fff, MWA_RAM },
  { 0x2000, 0x2000, wpcs_rombank_w }, /* 2000-23ff */
  { 0x2400, 0x2401, DCS89_YM2151_word_0_w }, /* 2400-27fe even *//* 2401-27ff odd */
  { 0x2800, 0x2800, DAC_0_data_w }, /* 2800-2bff */
  { 0x2c00, 0x2c00, hc55516_0_clock_set_w },  /* 2c00-2fff */
  { 0x3400, 0x3400, hc55516_0_digit_clock_clear_w }, /* 3400-37ff */
  { 0x3800, 0x3800, wpcs_volume_w }, /* 3800-3bff */
  { 0x3c00, 0x3c00, wpcs_latch_w },  /* 3c00-3fff */
MEMORY_END

________________________________

mjrgh:
Oh, small addition to the code snippet: you also need to add a new member, data8_t ym_reg_addr, to the static struct { } locals at line 909.  That will become apparent from the compiler error if you try to plug in the code and compile it.

Regards,
Mike

________________________________

droscoe:
Great description of the problem and solution!  I’m relatively new to this domain.  I’m a software engineer w/ almost 30 years in the field.  Started w/6502 assembly back in the 80’s.  Would love to get more into the hardware emulation side.  You mention documentation on some of these ROMs?  Could you point me to an example of what that looks like?  I would greatly appreciate it.

- Dave (Discord: Mr_Grynch#1235)

mjrgh:
> You mention documentation on some of these ROMs? Could you point
> me to an example of what that looks like?

You mean documentation for chips like the YM2151, I'm assuming.  That particular chip has at least a data sheet (the typical IC sheet with pin assignments and voltages and signal timing constraints and so forth) and something along the lines of a user guide/application note with some more details on the register layouts and music features.  Googling "YM2151 data sheet" and "YM2151 application manual" will turn up links.  There's probably at least as much information on that particular chip out there in various third-party projects - when I was chasing down the HC55516 interaction a few years ago, I came across a couple of those, one in particular was someone's FPGA re-creation of the chip, which had some useful insights into the internal architecture that are hard to piece together from the published documentation.  "YM2151 FPGA" turns up a couple of links there.

Regards,
Mike

________________________________

droscoe:
@mjrgh Thank you!

toxieainc:
Thanks, will have a look at this when i'm back on the weekend, i guess..
Would it make sense to file these observations as issue for YMFM? Maybe Aaron could just include this into the core itself?

mjrgh:
> Would it make sense to file these observations as issue for
> YMFM? Maybe Aaron could just include this into the core itself?

That would definitely be the more conceptually correct place to put it - it would make the YMFM emulator a more accurate YM2171 emulation, and would make the PinMame code a more straightforward 1:1 mapping to the hardware.  (In that the physical WPC89 boards didn't need some extra circuitry hacked onto the side to implement the delay timer, the way we need extra code hacked onto the side to implement it - the clock line latency was just an intrinsic property of the YM2171 chip's internals.)  So if Aaron's interested in implementing it there, that would be great.

On the other hand, handling it in wmssnd.c might be a more stable solution long-term.  The whole reason this came up again is that we fixed it in the "right" place the first time.  Since the fix was in the YM emulator, it broke as soon as we switched to a different YM emulator.  It'll break again in every future emulator upgrade/switch for the same reason - it's an undocumented aspect of the chip that no one would ever know about if it weren't for the dependency the Williams guys created when they wrote the WPC89 software.  Even if we never upgrade YM emulators again, I'd bet it'll still break every two years because someone accidentally or on purpose breaks it in the YMFM code base, in the course of fixing other bugs, or because they notice the weird and inexplicable complication of the latency timer and decide it needs cleaning up.  And of course, when it does come up again on the two-year cycle, we're all going to have forgotten the details, even if we vaguely remember the locus of the issue, so someone will have to spend a couple of days hunting it down again.

So, I don't know, both approaches have their merits and drawbacks.  As much as I think it's more righteous to fix it in the YM code, I think it might be more practical to use the wmssnd.c fix.

Mike

________________________________

toxieainc:
Filed as https://github.com/aaronsgiles/ymfm/issues/52

toxieainc:
I commited the YMFM timing/delay workaround now, thanks! Will also merge the PR, of course.

Thanks for the great work (as usual :))!

PR #146, opened 2023-08-24
Roscoe altsound bugfix 20230823

droscoe:
Incorporated review comments, removed commented code.  Added BSD-3 banners


jsm174:
Just throwing this out there, but would it be possible to start to doing some rebases instead of merges? 

The codebase now has 16 of these "Merge branch 'master'" commits in the repo after the AltSound PR, which makes it a little confusing when walking through history.

droscoe:
> Just throwing this out there, but would it be possible to start to doing some rebases instead of merges?
> 
> The codebase now has 16 of these "Merge branch 'master'" commits in the repo after the AltSound PR, which makes it a little confusing when walking through history.

Sure, I can set it to rebase instead

toxieainc:
Perfect, thanks!!

Issue #126, opened 2023-05-31
Vpinmame X64 hangs if using altsound 2 and Pinsound is not running


alaw-lufc:
(same issue as https://github.com/vpinball/pinmame/issues/81)

VPinMAME-sc-3.6-450-9d295cf-win-x64 will hang on startup if altsound mode is set to 2 but Pinsound is not running.

When Pinsound Studio 18.8.3 is running vpinmame X64 altsound 2 can still work as seen in the log part of the Pinsound window below;
![image](https://github.com/vpinball/pinmame/assets/73988577/d3ac9e76-5209-4f0a-a4bb-cbab62af23e5)
![image](https://github.com/vpinball/pinmame/assets/73988577/ada69100-07de-4dfc-85e7-ef4b206b2a6e)

This behaviour will also cause VPX X64 to hang at "starting game scripts" if Pinsound is not running. Ideally the game would still launch as it does in 32bit and use rom sounds when Pinsound is not running.

(a copy of Pinsound 18.8.3 is available here if that helps [https://www.dropbox.com/s/n66z7ylpwzld1uw/PinSoundStudio-windows-18.8.3.zip?dl=0](url))

toxieainc:
@droscoe Could you maybe have a look why this hangs? As you're revisiting the other altsound path at the moment anyway.  :)

droscoe:
> (same issue as #81)
> 
> VPinMAME-sc-3.6-450-9d295cf-win-x64 will hang on startup if altsound mode is set to 2 but Pinsound is not running.
@alaw-lufc  Can you provide the steps you used to show this problem.  I did a quick test using your initial description and did not reproduce the issue.  Thanks!

droscoe:
UPDATE:
I was able to reproduce the issue.  I can't get a proper stack trace from VS.  Says all running threads are external processes.  Here's what I see:
```
PinSound: Cannot open slot: \\.\mailslot\PinSoundStudio. Error: 2.
PinSound: Communication read slot opened: \\.\mailslot\PinSoundPinMAME.
PinSound: WriteSlot failed with 2.
```
Seems it's stuck waiting on acknowledgement from PinSound, and doesn't time out.

droscoe:
I was able to reproduce the issue.  However, I couldn't get a proper stack trace.  VS complained that all running threads were external.  The log showed the following:
```
NMI RATE SET TO 248
set visible area 0-319 0-41
Physical width 322, height 202
[...]

PR #154, opened 2023-09-01
Fixed hang if using PinSound and Studio is not running

droscoe:
This PR addresses #126 

The problem was with opening a write mail slot .  This obviously fails if PinSound Studio is not running, but we were not checking this.  We then proceed to try to open a read mail slot with a bad file handle.  This causes the code to hang in 64-bit.  By checking for the bad handle and aborting the function, the problem is resolved.


toxieainc:
Thanks, as usual!  :)

PR #184, opened 2023-12-30
PPUC should use libppuc, libzedmd, ...

mkalkbrenner:
I wanted to finish PPUC first before creating the next PR. But the builds fail since the latest changes to alphanummeric displays. @toxieainc fixed those in libpiname, but not in ppuc.
So it makes sense to merge the current state to get the builds back to green.

Regarding PPUC, a lot of stuff is working in that stage: lights, sound, switches. Coils need some debugging. But in general. all of these details are now handled in libppuc as an external dependency. So the PR is ready to be merged.

Also the hardware reached the testing stage:
https://www.youtube.com/watch?v=RBCEf7PU0Cs


toxieainc:
Great, thanks for the fixes!

PR #186, opened 2024-01-07
PWM: new implementation

vbousquet:
This is a rewrite of the preliminary PWM support I added at the beginning of the 3.6 phase. While using it, quite a few limitations arise, so I rewrote it to:
- Supports variable rate data writing/integration (needed for fast hardware with <1ms strobes)
- Supports high frequency controller using a linear PWM preintegration applied on high frequency data before physics model (needed for fast hardware with <1ms strobes)
- Supports strobed/PWM lamp matrix (Stern Whitestar and SAM uses this a lot, but it also improves the rendering on other hardwares)
- Supports Alphanumeric segment dimming (not yet exposed to VPinMame)
- Extended to all GI & Solenoids outputs, including aux board (WPC's GI really benefits from this, flashers wired on aux board  have been missing)
- Allow output definition from PinMame driver as well as from client app (allow mods & variants, also allow to support more hardware configuration)

In turn, the rewrite led me to create a sort fo generalized physic output support, then port to it, trying to unify drivers around it. This led to remove (after porting them to the new physic output system) the following implementation:
- the initial modulated solenoids,
- the preliminary PWM introduced in 3.6
- RGB lamps (which were only used by SAM driver to output linearly modulated lights)

This commits includes implementations for the following drivers:
- WPC,
- S11,
- Sega/Stern Whitestar,
- Stern S.A.M.,
- Capcom,
- GTS3

While testing, it appeared that timings were wrong for Capcom hardware, so this commits also contains an improved Capcom driver with better timings (better U16 emulation, fix to MC68k core,...), needed to get correct lamp fading.

This commits also contains the physic hardware definition for the following games: all Capcoms, all WPC, most SAM, GNR for S11, LOTR for Whitestar, none of the GTS3.

The aim would be to have all the physic device connected to binary outputs defined on the PinMame side but it is too much work to be done straight away. Since the implementation allows to define physics device at runtime (for mods & variants), this also allows to define the missing hardware directly from the user application if we do not gather all the info for 3.6.

Using the new implementation is fairly straightforward (at least for games with hardware already defined): just set SolMask(2) = 6, then changed lamps / solenoids / gi will be 0..255 value relating to the main characteristic of the output (relative brightness for bulbs & LEDs).

Since this is a deep change, I think there will be bugs I jave missed, so this would benefit for large user testing. Also, @mkalkbrenner @jsm174 I tried to keep things clean for PPUC and LibPinMame (and PROC) but since I don't use them, the testing is fairly low.

### DIFF
--- a/src/cpu/m68000/m68kmame.c
+++ b/src/cpu/m68000/m68kmame.c
@@ -1325,7 +1325,7 @@ static void m68306_duart_set_dusr(int which,int data);		//Set the DUSR register
 static void m68306_duart_set_duisr(int data);				//Set the DUISR register
 static void m68306_duart_set_duimr(int data);				//Set the DUIMR register
 static void m68306_duart_check_int(void);					//Check for a DUART interrupt
-static void duart_start_timer(void);						//Duart Start Timer
+static void duart_start_timer(int reset);						//Duart Start Timer
 static void duart_timer_callback (int param);				//Duart Timer Callback
 static void m68306_rx_cause_int(int which,int data);		//Generate a Receiver Interrupt
 #ifndef PINMAME_NO_UNUSED	// currently unused function (GCC 3.4)
@@ -1525,22 +1525,15 @@ static data16_t m68306_duart_reg_r(offs_t address, int word) {
 		case 0xf7fd:
 			LOG(("%8x:START COUNTER COMMAND Read = %x\n",activecpu_get_pc(),data));
 #if !DISABLE_68306_TIMER
-			//TIMER MODE?
-			if(m68306duartreg[dirDUACR] & 0x40) {
-				//Look for 1->0 transition in the output to trigger an interrupt!
-				if(m68306_duart.timer_output) {
-					//Set CTR/TMR RDY bit (3) in DUISR
-					m68306_duart_set_duisr(m68306duartreg[dirDUISR] | 0x08);
-				}
-				//Clear OP3 output
-				m68306_duart.timer_output = 0;
-				//Restart timer!
-				duart_start_timer();
+			//Look for 1->0 transition in the output to trigger an interrupt!
+			if(m68306_duart.timer_output) {
+				//Set CTR/TMR RDY bit (3) in DUISR
+				m68306_duart_set_duisr(m68306duartreg[dirDUISR] | 0x08);
 			}
-			//COUNTER MODE?
-			else {
-				//Start count down from preloaded value
-			}
+			//Clear OP3 output
+			m68306_duart.timer_output = 0;
+			//Start timer!
+			duart_start_timer(1);
 #endif
 			break;
 		//F7FF - STOP COUNTER COMMAND
@@ -1550,11 +1543,12 @@ static data16_t m68306_duart_reg_r(offs_t address, int word) {
 			//Clear CTR/TMR RDY bit (3) in DUISR - BOTH TIMER & COUNTER MODE
 			m68306duartreg[dirDUISR] &= (~0x08);
 
-			//COUNTER MODE? - NOT IMPLEMENTED
+			//COUNTER MODE
 			if((m68306duartreg[dirDUACR] & 0x40) == 0) {
 				//Clear OP3 output
 				m68306_duart.timer_output = 0;
-				//Stop the counter..
+				//Stop the counter
+				timer_enable(m68306_duart.timer, 0);
 			}
 #endif
 			break;
@@ -1819,7 +1813,7 @@ static void m68306_rx_cause_int(int which,int data)
 }
 
 //Duart Start Timer
-static void duart_start_timer(void)
+static void duart_start_timer(int reset)
 {
 	double time;
 	double clock_src;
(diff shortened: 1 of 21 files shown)


jsm174:
Wow, this is amazing! From the looks of it, I think all I need to do is modify 

https://github.com/vpinball/vpinball/blob/66942f17ceebf2ff0badec6d94431cf12a584f2d/standalone/inc/vpinmame/VPinMAMEController.cpp#L1071-L1089

and I should be good on the standalone side.

vbousquet:
There is another thing to be consider for LibPinMame: you are using OnSolenoid events. I just ported the code for it but I think this could be adapted to benefit for the change, for example by directly processing events for SOL_2_STATE while delaying the other to VBlank (as they are now). This should improve the latency quite a lot if event are directly processed

vbousquet:
I forgot to share the updated core.vbs for VPX that ease using this (I will commit it to VPX when VPM will have th efeature integrated):
[core.zip](https://github.com/vpinball/pinmame/files/13854767/core.zip)

toxieainc:
Great!
One more thing: Could you please generalize the game name detections some more? Otherwise just the newest ROM revisions will work correctly i guess..

toxieainc:
..oh, and some text for the whatsnews, please.. Thanks!

vbousquet:
Whatsnew update (directly in commit, I force pushed it)
The game name detection is already generalized: it walk up to the root game name and compares against it, therefore handling all the clones at once.

vbousquet:
Ooopss, you were fast :) my force p ush was after the merge.
I will do a new PR for the whatsnew

toxieainc:
> The game name detection is already generalized: it walk up to the root game name and compares against it, therefore handling all the clones at once.

Really? Only GTS3 seems to be, but what about these?
https://github.com/vpinball/pinmame/blob/419c3360fbd0b10bd17384382118d65d921cb97e/src/wpc/s11.c#L818
https://github.com/vpinball/pinmame/blob/419c3360fbd0b10bd17384382118d65d921cb97e/src/wpc/wpc.c#L1176
https://github.com/vpinball/pinmame/blob/419c3360fbd0b10bd17384382118d65d921cb97e/src/wpc/capcom.c#L735
https://github.com/vpinball/pinmame/blob/419c3360fbd0b10bd17384382118d65d921cb97e/src/wpc/sam.c#L1250

volkenborn:
Hi guys,

this change in core.c will lead to a lamp matrix 8 times the size as it 
was before:

-    for (ii = 0; ii < CORE_CUSTLAMPCOL + core_gameData->hw.lampCol; ii++) {

+    int nCols = coreGlobals.nLamps ? (coreGlobals.nLamps + 7) >> 3 : 
CORE_CUSTLAMPCOL + core_gameData->hw.lampCol * 8;

Was that intentional, or is it a mistake (actually I hope for the latter)???


Cheers,
Gerrit


Am 08.01.2024 um 09:41 schrieb toxie:

vbousquet:
> Hi guys, this change in core.c will lead to a lamp matrix 8 times the size as it was before

It's a bug! Thanks for spotting, I will remove the `* 8` in ```int nCols = coreGlobals.nLamps ? (coreGlobals.nLamps + 7) >> 3 : CORE_CUSTLAMPCOL + core_gameData->hw.lampCol * 8;```

toxieainc:
> > > The game name detection is already generalized: it walk up to the root game name and compares against it, therefore handling all the clones at once.
> > 
> > 
> > Really? Only GTS3 seems to be, but what about these?
> 
> I may have miss something. The aim of this code (which is present for all drivers) is to move to the root gamedef:
> 
> ```
>   const struct GameDriver* rootDrv = Machine->gamedrv;
>   while (rootDrv->clone_of && (rootDrv->clone_of->flags & NOT_A_DRIVER) == 0)
>      rootDrv = rootDrv->clone_of;
>   const char* const gn = rootDrv->name;
> ```
> 
> Then I use the name of the root game definition. I don't see the difference with GTS3.

Whoops, you're completely right. Sorry for the noise..  :/
Can you then please commit the lamp matrix change, cause then i could start the next PinMAME beta.. Thanks!

toxieainc:
Oh my, seems my brain is still in xmas mode  ;)
Thanks!

Then i'll also wait some more, cause this sounds like an extremely cool feature to have..

mkalkbrenner:
@vbousquet thanks for working on this and for pinging me.
I'll test things with PPUC, but it will take some time because we focused on WPC during the last months. To test SAM we need to "configure & wire" a test case first.
@toxieainc so don't wait for my feedback and go ahead.

vbousquet:
> I'll test things with PPUC, but it will take some time because we focused on WPC during the last months. 

The new implementation replace modulated solenoids that were implemented for WPC and SAM, so **it changes WPC as well**.
Note that it also adds an integrator that may be of interest for PPUC to drive LEDs from old controller logic.

jsm174:
@vbousquet, for vpx standalone, there's a runner thread, but everything else should work like the normal controller. 

So I'm guessing I will need to just call  `core_request_pwm_output_update()` in 

https://github.com/vpinball/pinmame/blob/505d906a1f6d113f646007c85eddd8f32bc06cb8/src/libpinmame/libpinmame.cpp#L1035C16-L1038

and do the same for all the other functions?

toxieainc:
Great stuff!
What do you recommend as API for VPM?

vbousquet:
> So I'm guessing I will need to just call `core_request_pwm_output_update()` in....

Yes, this is what schedules the update. So all calls coming from another thread should call it (it is thread safe).

This design has the benefit of nearly 0 synchronization cost but the drawback that the data returned is always late (the one updated asynchronously from the last request, except for outputs that do not perform integration at all like 2 state solenoids). If the update request is called regularly, this may lead to not perfectly smooth fading. I tested it and it never happened to me so I kept this since it lower the CPU load (16 times lower) but if we find issue, we can get back to the systematic update every 1ms (the CPU load with this is still around 5% for a 10 year old computer so... but with the alpha 640 segments of a GTS3 around 9% of the CPU time is spent in the output integrators).

> Great stuff!

Thanks! this needs to be merged though: https://github.com/vpinball/pinmame/pull/190

> What do you recommend as API for VPM?

My feeling is that the arch and the API needs to have a global evolution since the way we interface VPX and VPM has the following drawbacks from my point of view:
- it creates high latency (no callbacks, B2S in the way using Windows registry for communication,...)
- it uses low precision bytes where floats would be needed
- it freezes the API and hinder progresses since every evolution of VPM needs an evolution of VPX, B2S and often tables, all of this needing to be more or less coordinated

So, we have a few options here:
- either do the full thing: create the module/plugin arch that will allow to normalize standalone modules as plugins/modules on all platform, move VPM to one of these using LibPinMame, use standalone B2S implemntation as one of these plugins, keep B2S (or a module) just for its plugin, or even just port the plugins to the new plugin... We could also have a controller object, hiding the API and allowing easier evolution (to avoid tables needing to be updated). This is huge and will take a looonnngggg time, so definitely not for 10.8
- don't do anything (so no dimmed segments)
- change the existing API call to return different values depending on physics output being enabled or not => this would need B2S to be updated since it processes this data as well => not a good idea since it makes the API ocmplex and needs VPX+VPM+B2S update
- create a new API call, update B2S for it & update core.vbs => this can be done, but this will postpone one more time 10.8 (not a real problem, but at some point, it would be nice to move on) and will create an API call that will be deprecated if we do the full ting

volkenborn:
Hi again,

now in the console version all solenoids numbers logged to screen have a value 1 less, so #1 will write out "0", #2 will write "1", and so on.

This is due to a change in core.c, lines 1500 and 1532 (ii starts with a value of 0, before it was 1).
I suppose we can change it back to 1, and call OnSolenoid without the extra increment?!

PR #194, opened 2024-01-17
Generalized pwm

vbousquet:
A new bunch of fixes for the new physic outputs.


toxieainc:
Great stuff as usual! Curious to test the dimmed segments  :)
Could you please also update the whatsnew?

PR #200, opened 2024-01-24
Fix erroneous application of ducking for music samples

droscoe:
Normal convention is to set the ducking value for music samples on channel 0 to 100, since music should not duck music in traditional altsound packages.  However, setting it to anything other than 100 will cause music to duck other music samples.  Setting it to 0 will completely mute other music samples, which is undesirable.  This fix will force music sample ducking to 100, ensuring proper operation.


toxieainc:
Great, thanks!!

PR #203, opened 2024-01-25
feat: sdl2

francisdb:
I ported the SDL code to SDL2.

X11 is ancient tech that is going away.
`xpinmame` on linux now builds against SDL2 for graphics and sound.

Some light crackling on the sound, probably something to do with pinmame not coming up with audio data in time?

```
SDL_AUDIODRIVER=pipewire SDL_VIDEODRIVER=wayland ./build/xpinmame -rompath ~/.pinmame/roms -nvram_directory ~/.pinmame/nvram  t2_l8
```

image

No luck with hidpi for now

### DIFF
--- a/.github/workflows/xpinmame.yml
+++ b/.github/workflows/xpinmame.yml
@@ -62,7 +62,7 @@ jobs:
 
       - if: matrix.os == 'ubuntu-latest'
         run: |    
-          sudo apt install libx11-dev libxv-dev libasound2-dev sdl2-dev
+          sudo apt install libx11-dev libxv-dev libasound2-dev libsdl2-dev
       - name: Build xpinmame-${{ matrix.platform }}
         run: |
           cp cmake/xpinmame/CMakeLists_${{ matrix.platform }}.txt CMakeLists.txt
--- a/.github/workflows/xpinmame.yml
+++ b/.github/workflows/xpinmame.yml
@@ -61,8 +61,8 @@ jobs:
           brew install sdl2
 
       - if: matrix.os == 'ubuntu-latest'
-        run: |    
-          sudo apt install libasound2-dev libsdl2-dev
+        run: |
+          sudo apt update && sudo apt install libasound2-dev libsdl2-dev
       - name: Build xpinmame-${{ matrix.platform }}
         run: |
           cp cmake/xpinmame/CMakeLists_${{ matrix.platform }}.txt CMakeLists.txt
--- a/.github/workflows/xpinmame.yml
+++ b/.github/workflows/xpinmame.yml
@@ -62,7 +62,7 @@ jobs:
 
       - if: matrix.os == 'ubuntu-latest'
         run: |    
-          sudo apt install libx11-dev libxv-dev libasound2-dev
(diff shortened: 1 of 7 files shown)


francisdb:
Questions:

* is it ok to enable sdl2 by default on Linux?
* do we also want to do this for macOS?

francisdb:
Nobody interested?
Is this a bad idea?
Some feedback would be nice.

jsm174:
The more builds and the more options the better in my opinion.

It would be nice to get a true MacOS build going without needing to install XQuartz.

Is this PR ready to be merged? If not could it be switched to draft?

jsm174:
> No idea how to fix this inline stuff, no issues here on my machine. Must be some clang 14 vs 15 or arm vs x64 issue?

If you look at the other cmakes, we had to do this:

`add_definitions( "-DINLINE=static inline __attribute__((always_inline))" )`

So you remove the entry you have in `add_compile_definitions` ie, ` INLINE=static __inline` and use above?

francisdb:
> > No idea how to fix this inline stuff, no issues here on my machine. Must be some clang 14 vs 15 or arm vs x64 issue?
> 
> If you look at the other cmakes, we had to do this:
> 
> `add_definitions( "-DINLINE=static inline __attribute__((always_inline))" )`
> 
> So you remove the entry you have in `add_compile_definitions` ie, ` INLINE=static __inline` and use above?

Thanks, tried a different version I found on `vpinball` that did not work. What a mess.

jsm174:
> Few remaining questions:
> 
> * Do I remove the commented parts in the cmake files (x11)?

I'd defer on this question. Maybe if X is no longer needed, it gets converted to just pinmame?

> * How/where do we document that sdl2 is required for xpinmame on mac / linux?

If you compiled with the static libSDL2.a would you even need SDL2 as a requirement? It would be directly in the binary.

> * Do I duplicate `CMakeLists_osx-x64.txt` to `CMakeLists_osx-arm64.txt` or remove the platform prefix as the current x64 works for both?

For now I would keep multiple CMakeLists.txt. I understand there is one line different, but I think we need an overhaul of CMake, and I'd prefer to do that in one shot.

francisdb:
> I seem to use a different include of sdl in my sdl2 version. Do you know why you inlcude SDL2/SDL.h and I include just SDL.h

No idea, but does not seem to be an issue on the github builds.

gnulnulf:
I used makefile.unixsdl which contained sdl-config, after changing that to sdl2-config and SOUND_SDL=1 it worked on my side.
The clicking sound could be a signed/unsigned issue. Sound in mpu35 is good.

gnulnulf:
From my side no progress, I was able to do my rom things with the current sdl2. Maybe a nice winter project. Are there any volunteers to migrate directX7 to SDL3?

francisdb:
> Hi! As this has been hanging around for a while, but it seems useful.. I'm no Linux expert, so have no opinion, plus by now it could make sense to directly go to SDL3.. How to progress here then? Thanks!

I might have a go at SDL2->SDL3

PR #218, opened 2024-02-07
Added altsound command processing support for Gottlieb System 80A

droscoe:
There is currently no support for altsound for Gottlieb games.  This became apparent with the recent release of an altsound for Alien Star (Gottlieb 1984).

I've added basic support for Gottlieb System 80A, so users can create altsounds for them.  I have confirmed that my code solves the issues seen with the Alien Star altsound package.


toxieainc:
Thanks!!

PR #245, opened 2024-03-10
Add build number and SHA7 to the file properties

JockeJarre:
I added the build number as well as the sha7 hash value to the file properties and made it analogue to how it is done in Visual Pinball:

Git Workflow updates git_version.h with both values and has version.h include this file and model different types of version strings.
I would have put the git_version.h and version.h in the root folder so that it can be used globally, but I do not know how you want it?

image

VPinMame automatically returned the build number in the COM version method, but it got it wrong:

The version string went from 03060000 to 0306000870 (one character more).  
So I hardcoded 2 zeros instead keeping 03060000 for now. I will address this issue separately.

One more thing to think about, if you now throw in a new VPinMAME.dll and start the Setup routine, it says that you have the file which you are about to install already installed... It cannot find out the right version since it was already replaced.

image

This might need to be rephrased?


toxieainc:
Thanks, i guess this is all good, but as you asked: Please put the version files into root, then we can use it everywhere more easily, thanks!

JockeJarre (commented):
Thanks @toxieainc , now the files are moved to the root.

toxieainc:
Thanks!

rockfordroeNG:
This is causing me some compilation errors when building with the outputs of "create_vc2002_from_vc2012.bat".

PR #252, opened 2024-03-13
redo get version from version.h

JockeJarre:
Hi @toxieainc, I finally managed to redo the "get version from version.h".

WOW, you brought me into cmake hell ;) I have learnt a lot about cmake and regular expressions now...

The "problem" was that the version.c was referenced in many places and read by cmake and in Github actions.
 
I let them stay as 3.6 where it is still needed, otherwise we now use 3.6.0.4711, where possible.

Will now continue with the COM methods.


toxieainc:
Cool, thanks a lot for going through that pain!!

Issue #196, opened 2024-01-18
Flipper issues on WOF (was: Lamps above 400 on Wheel of Fortune)


francisdb:
First, is there a better place to ask these questions?

This table used to play without issues but `Controller.ChangedLamps` now seems to return lamps over 400 where the table code has only 400 entries in the arrays. Is this expected as it might break more tables?

```
2024-01-18 09:46:07.837 INFO  [28830] [DebuggerModule::Print@4232] Script.Print 'lamp 580 state 255'
2024-01-18 09:46:07.847 INFO  [28830] [DebuggerModule::Print@4232] Script.Print 'lamp 583 state 255'
2024-01-18 09:46:07.866 INFO  [28830] [DebuggerModule::Print@4232] Script.Print 'lamp 581 state 255'
2024-01-18 09:46:07.886 INFO  [28830] [DebuggerModule::Print@4232] Script.Print 'lamp 582 state 255'
2024-01-18 09:46:11.757 INFO  [28830] [DebuggerModule::Print@4232] Script.Print 'lamp 580 state 0'
2024-01-18 09:46:11.788 INFO  [28830] [DebuggerModule::Print@4232] Script.Print 'lamp 583 state 0'
2024-01-18 09:46:11.916 INFO  [28830] [DebuggerModule::Print@4232] Script.Print 'lamp 581 state 0'
2024-01-18 09:46:12.167 INFO  [28830] [DebuggerModule::Print@4232] Script.Print 'lamp 582 state 0'
2024-01-18 09:46:40.307 INFO  [28830] [DebuggerModule::Print@4232] Script.Print 'lamp 601 state 0'
2024-01-18 09:46:36.137 INFO  [28830] [DebuggerModule::Print@4232] Script.Print 'lamp 609 state 255'
2024-01-18 09:46:40.317 INFO  [28830] [DebuggerModule::Print@4232] Script.Print 'lamp 595 state 247'
2024-01-18 09:46:40.338 INFO  [28830] [DebuggerModule::Print@4232] Script.Print 'lamp 595 state 121'
2024-01-18 09:46:40.357 INFO  [28830] [DebuggerModule::Print@4232] Script.Print 'lamp 588 state 0'
```

PINMAME_SHA beec0c3cda92d3df1caeeb8af83ef14b05e1cf06

vbousquet:
This is intended and is only for WOF and WPT. It happens that these tables have LED matrices which seem to be PWMed so I added them as such.

[...]

PR #255, opened 2024-03-16
Fix WOF overlay between lamps & solenoids

vbousquet:
Fix https://github.com/vpinball/pinmame/issues/196


francisdb:
Great to see this one finally fixed. Does this have any influence on the script?

Issue #249, opened 2024-03-11
Add new COM Version method including build number as double


JockeJarre:
Similar to the issue on VPX: https://github.com/vpinball/vpinball/issues/1558

**Is your feature request related to a problem? Please describe.**
To help the flipper table creators, not having to tell in all forums which minimum version to use for VPX, VPM and B2S

**Describe the solution you'd like**
Currently the Version method **returns a string** as "03060000" for VPM 3.6, it would be an improvement if the build number would be added as decimal value like "03060000.0879" (allowing 4 decimal characters), meaning one would have to change the method to return a bigger string instead. 

**Describe alternatives you've considered**
I have been thinking of using a separate Version method. But rather I would keep the same.  I also considered changing the datatype to double as in VPX, but then a lot of scripts and tables would have to be changed.

**Additional context**
I have tested the current usage of controller.FullVersion in the local vbs files, and the same if statements still work perfect.

Some examples from "core.vbs":

1.  If Controller.Version >= "01200000" Then
3.  If Controller.Version >= 03060000 Then      **one note, the correct usage would be to use "03060000" comparing two strings!**

Using this it will be possible to test using the build number as decimal:
1.  Controller.Version >= 03060000.0879    is True  (where Controller.Version is a string with value "03060000.0880")

The same also works in older versions
1.  Controller.Version >= "03060000.0879"  is False (where Controller.Version is an string with value "03060000")

toxieainc:
@JockeJarre While at it, seems i screwed up the actions via my additional version changes.
But i don't get it, as things build locally for me.
Could you please check https://github.com/vpinball/pinmame/commit/a4a9c2232eb9b3e4bcb53361a031afc0c4ba94ea what i did wrong? Thanks!

JockeJarre:
Will do!

JockeJarre:
I found out what the problem is. I have reverted src/version.c for now.
The workflow tasks are reading out the "3.6" from version.c and build version strings from that.

I will clean it up within this issue and put your changes in. 
**I do not think I can push directly to this repo, so I will make a pull request.**

JockeJarre:
@toxieainc Do you know if and how to add sourcecode to the *_VC20XX.vcxproj.filters and *_VC20XX.vcxproj.filters files?
Is there any common logic to follow? I think I added something wrong in last issue which got merged.

toxieainc:
Will look into it..

toxieainc:
[...]

PR #257, opened 2024-03-17
Add build number to the COM Version method of VPM #249

JockeJarre:
Currently the Version method returns a string as "03060000" for VPM 3.6, it would be an improvement if the build number would be added as decimal value like "03060000.0879" (allowing 4 decimal characters), meaning one would have to change the method to return a bigger string instead.


toxieainc:
Thanks! But should we go directly to 5 digits? 4 seems like it could run out in some years.

JockeJarre:
@toxieainc, I am discussing with @vbousquet if this ticket is such a good idea.  Continued on VPU...

JockeJarre:
We concluded that it isn't good to include the build number in the COM methods, as it would encourage more usage of the builds rather than the official releases.

Issue #251, opened 2024-03-13
VPinMAME 3.6-896 many problems like crashes on start, blown out lamps, no starts


passionforpins:
Ran ACDC LUCI (Stern 2013) VPW Mod 1.0.2 with vpx 10.8.0 RC v1923 x64 dx and vpinmame -sc 3.6-896 x64 and crashed on start with error shown attached.  I rolled back to vpinmame 888 and all ok.

Ran Last Action Hero (Data East 1993) VPW v1.2 with same as above but pressing start no ball enters the shooter lane

Ran Iron Man Vault Edition (Stern 2010) VPW v1.0 with same as above but nearly impossible to play as all lamps are blown up
![acdc-luci-vpw-102-vpinmame-crash](https://github.com/vpinball/pinmame/assets/52011304/8a6ce8e6-d506-4f3a-91c9-645ab8e236db)

toxieainc:
Somehow i cannot open the link. It just displays nothing.

mazjohn:
I posted this in another closed action, so wasn't sure if it would get read- otherwise, so including my feedback here since seems to be on topic -  After some more extensive testing of VpinMame build 896 combined with VPX X64 build 923, I have also concluded that there are many tables experiencing crashes during load or after starting a game with ball 1. An example of this is with VPW's ACDC Luci table V1.1, where it throws Runtime error
------------- Line: 4202 Overflow: '[number: 34680]' after shooting ball 1- -- I've reverted back to VpinMame build 720 with the same latest VPX X64 build 923, and the issues go away. It seems in all of my testing for backward compatibility, I conclude that if backward compatibility was my primary concern, that the last usable beta build of VpinMame for me was build 720 dated Dec31,2023. That build has no flasher issues, table crashes, or overblown lighting... and where I am holding at for time being.  Thank you-

toxieainc:
Thanks for the tests.
The crashes and blown out lamps,etc are all due to my latest PWM changes i guess. So just be patient, this will get fixed.

studlygoorite:
Just adding that v906 still has this issue, with VPW's X-Men so far anyway.

apophis79:
@toxieainc In the case of VPW X-Men, it is set up with `UseVPMModSol = True` which should not include insert lamp fading... only `UseVPMModSol = 2` should do that. However `Controller.ChangedLamps` is returning modulated values from 0 to 255 in this table. The table expects the returned values to be either a 0 or a 1.

[...]

PR #258, opened 2024-03-17
PWM: Fix internal flag being used for data exchange

vbousquet:
Fix https://github.com/vpinball/pinmame/issues/251


toxieainc:
Thanks a lot!!!

Issue #268, opened 2024-04-23
Hickup on AFM when destroying saucer in VPinmame after v.888


Regulations:
I have trouble with all the versions of VPinmame after "VPinMAME-sc-3.6-835-60613e8-win-x64.zip" with Attack From Mars G5K and afm_113 rom. Using latest VPX 10.8 GL. Everytime I destroy a ship, right when the explosion happens there is a small hickup in the DMD, both visual and sound. It also happens if I use ColorDMD. 

See video:
https://github.com/vpinball/pinmame/assets/31146448/c0e87685-4e73-4fa4-a961-0c76deeb4687

toxieainc:
@Regulations do you use an altsound package or rom/emulated sound?

Regulations:
> @Regulations do you use an altsound package or rom/emulated sound?

I'm using rom sound. 
Alt. Sound Mode=0
![afm_113b](https://github.com/vpinball/pinmame/assets/31146448/5b272e21-698e-44cb-a886-bdb2d8f082a3)

Kraken74:
Here to confirm this bug.  Tested with GL and regular (DX?) VPX 10.8.0 RC3 and RC2.  Problem disappears with vPinMame 3.6-822.  Just a slight pause/hickup during the final explosion of the UFO and the UFO solenoid does't fire the final time.  Tested on two different tables and it happens on both.

https://github.com/vpinball/pinmame/assets/5919849/d729d807-2209-40c6-93ce-7fb8f4f6d654

Kraken74:
I noticed it yesterday while testing my colorization and you can easily see AND hear the glitch here:
https://youtu.be/TSkgCbfZDZk?t=72

eminence-afk:
noticed this on medieval madness, too. upon destroying a castle. (109c)

also with demolition man when the game resets after match feature. (h6b)

Loadedweapon:
Just here to confirm this issue.. Just a slight pause on latest vpinmame

sverrewl:
> noticed this on medieval madness, too. upon destroying a castle. (109c)
> 
> also with demolition man when the game resets after match feature. (h6b)

Good to see we at least see the same. I asked in VPW Discord about some of the oddities but no-one tried to verify my findings. Haven't tried demo-man, but, I can't get the draw bridge to drop for MM.

Loadedweapon:
The draw bridge is a script issue and has been posted

Loadedweapon:
[...]

Issue #266, opened 2024-04-17
newer pinmame, gives stutter in rom sound


sverrewl:
For the first time, couple of months back, I noticed that my cab, for quite a lot of table got some odd rom sound stutter. It could be almost any kind of sample being played, but, maybe most noticeable on pr. example Road Show, with Red's call-outs. She sounded like she got a stutter. I shrugged that off and though that it would be fixed in upcoming releases, but, it is still there.

I now wanted to investigate it further, is this a final blow for me, still running Win7 in the cab ?? Tried it on my Win10 desktop upstairs and had no issues. Until i checked the sound_mode set. Default upstairs is was on 0 and on cab 1. Changing it to 1 on Win10, I could reproduce this. Maybe not as easy to hear on the more beefier machine, but, I didn't have to play rs_l6 for long before samples seems to be played more than once in a row, so to speak. Not ending the first iteration.

The night we first noticed  this, was with some beers, so I didn't really take a note on which machines - but, we played many different ones from late 80 and early 90. Maybe WPC issue ?

toxieainc:
Hmmm.. You mean alt. sound mode = 1, right? 
If yes, this would be something for @droscoe

sverrewl:
Yes. Alt Sound Mode in the gui - sound_mode in registry. 1 used to be improved over 0, but, the "new" alt code kind of hi-jacked something that was working. I don't understand the reasoning for not making a new number for the new feature. Oh, well, I'm sure it gets fixed.

toxieainc:
In theory, existing sets should work more or less the same with the newer code. The code for these is not THAT different.
But it was also necessary to rewrite like this, my old/initial implementation was a complete mess.

So i guess its 'just' a bug.

droscoe:
> For the first time, couple of months back, I noticed that my cab, for quite a lot of table got some odd rom sound stutter. It could be almost any kind of sample being played, but, maybe most noticeable on pr. example Road Show, with Red's call-outs. She sounded like she got a stutter. I shrugged that off and though that it would be fixed in upcoming releases, but, it is still there.
> 
[...]

PR #270, opened 2024-04-29
PWM: improve MODSOL implementation

vbousquet:
- Core: Don't write lamp matrix PWM if not enabled
- Core: Fix integration not being done up to 'now'
- WPC: only write PWM data corresponding to the options
- also minor formating cleanups

Should improve https://github.com/vpinball/pinmame/issues/266 and https://github.com/vpinball/pinmame/issues/268


toxieainc:
Thanks!
But the part with 'now' is kinda fishy to me now. I did it like that to be able to make the integration more even/'smooth', i.e. with roughly same-length segments over which the PWM is integrated then. Otherwise one gets screw ups in the values, especially if its in a flicker fusion 'phase'.
Or was there some noticeable issue with some tables/machines with my code?

vbousquet:
I may not have understood the intent regarding the 'now' part. When I reviewed the code, it looked wrong to me because I think it would lead to integration of the wrong output value. Here is the original code:
```
  const float U = (((coreGlobals.binaryOutputState[index >> 3] >> (index & 7)) & 1) ^ output->state.bulb.isReversed) ? output->state.bulb.U : 0.f;
  const float dt_diff = (float)(output->state.bulb.integrationTimestamp - output->state.bulb.prevIntegrationTimestamp);

  if(U != output->state.bulb.prevIntegrationValue        // state flip?
     || dt_diff >= (float)(BULB_INTEGRATION_PERIOD*20.)) // or we waited long enough to get a stable discrete integration
  {
    // do the integration in a loop of small steps, roughly in BULB_INTEGRATION_PERIOD sized steps (but rounded up/down to have same sized cycles in here)
    float countf = floorf(dt_diff*(float)(1./BULB_INTEGRATION_PERIOD) + 0.5f);
    if(countf < 1.f) // single cycle/short pulses need to use the 'original' cycle time as a workaround
      countf = 1.f;
    const int count = (int)countf;
    const float dt = dt_diff/countf;
    for(int i = 0; i < count; ++i) {
      // Keeps T within the range of the LUT (between room temperature and melt down point)
      output->state.bulb.filament_temperature = output->state.bulb.filament_temperature < 293.0f ? 293.0f : output->state.bulb.filament_temperature > (float) BULB_T_MAX ? (float) BULB_T_MAX : output->state.bulb.filament_temperature;
      const float Ut = output->state.bulb.isAC ? (1.41421356f * sinf((float)(60.0 * 2.0 * PI) * (float)(output->state.bulb.prevIntegrationTimestamp - coreGlobals.lastACZeroCrossTimeStamp)) * output->state.bulb.prevIntegrationValue) : output->state.bulb.prevIntegrationValue;
      const float dT = dt * (float)bulb_heat_up_factor(output->state.bulb.bulb, output->state.bulb.filament_temperature, Ut, output->state.bulb.serial_R);
      output->state.bulb.filament_temperature += dT < 1000.0f ? dT : 1000.0f; // Limit initial current surge (1ms is a bit long when emulating this part of the heating)
      core_eye_flicker_fusion(output, (float)bulb_filament_temperature_to_emission(output->state.bulb.filament_temperature));
      output->state.bulb.prevIntegrationTimestamp += dt;
    }
    output->state.bulb.prevIntegrationTimestamp = output->state.bulb.integrationTimestamp;
    output->state.bulb.prevIntegrationValue = U;
  }
  // else: wait with the integration until some more time vanished or the state will flip
  output->state.bulb.integrationTimestamp = now;
```

When integration is performed, you set the next integration value to be done on `U` starting from `output->state.bulb.integrationTimestamp`, but at this point `output->state.bulb.integrationTimestamp` is pointing to the previous integration timestamp while `U` value is valid starting from `now` timestamp. This does not seem correct to me.

toxieainc:
Thanks, i will think through all that again, hopefully this evening.

toxieainc:
..sorry, still on my todo-list to double check this.

But i guess what confused you, is the overall delay of the integration.
In order to collect and process (roughly) same-sized time intervals to not screw with the PWM 'conversion' quality, there is always a delay by buffering things.
Due to the regular triggering of the updates, this delay is not that bad in practice though and should not be noticeable.
I also double checked the output of the final results back then (numeric values), and this looked all sane.

But maybe i still missed something.. .. ..

toxieainc:
..small update as i had some more time: I just double checked the implementation and the outputs again, all still looks fine to me with the current scheme

PR #277, opened 2024-05-28
Add timekeeper support to pinmame

Tiger21820:
I added the timekeeper files (don't know if it's all of them) from the main mame repository to the machine directory for pinmame if anyone wants to add timekeeper functionality so Stern Tournaments (ToPS) will work correctly!

### DIFF
--- a/src/machine/timehelp.h
+++ b/src/machine/timehelp.h
@@ -1,0 +1,58 @@
+// license:BSD-3-Clause
+// copyright-holders:Aaron Giles,smf
+/***************************************************************************
+
+    timehelp.h
+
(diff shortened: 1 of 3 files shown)


toxieainc:
But will this even compile like this?

Tiger21820:
I'm not really sure, so if this request cannot be processed, then I or someone else can close it, and if anyone wants to follow up on this, they can re-open or submit a new pull request!

toxieainc:
I will close this then. We know that the files exist already, so just adding them without wiring them up does not make too much sense. Sorry!

PR #281, opened 2024-06-15
Add external synchronization using a time fence

vbousquet:
The existing implementation uses subdivided vblank (virtual ones at 60Hz) to perform emulation synchronization. This works well but can lead to some latency when used with a physic emulator like Visual Pinball X.

For example, the following MSVC Concurrency Viewer shot shows the VPinMame thread (last one above DX GPU threads). The blue part is when VPinMame is ahead of time, therefore sleeping to catch up, the green dots shows frame ticks of VPX (on a 144Hz display).
image

The problem is that input and physics (leading to switch triggers) can happen during these sleeping periods (with some additional latency while waiting for synchronization between VPX & VPM). Example of these are nearby optos like in XMen or Stern's Star Trek, video modes like 'Walk the Plank' in Black Rose, or flipper bats controlled by the microcontroller (leading to the need of fastflip hack).

The proposed change allow PinMame to be synchronized externally against an externally controlled main clock giving a time fence that PinMame will honor. Using this (and the latest changes in VPX 10.8.1), the previous MSVC concurrency viewer become the following (VPinMame is the last Worker Thread, unlabelled green span at the bottom are VPX syncing to VPM, dots are frame marker):
image

Here, you see VPM catching up at the pace of VPX, keeping a maximum margin of 1ms between the two. Both of them are synchronizing in between, leading to better synchronized emulation.

Input latency in this scheme could theoretically be down to 1.5ms: average of 0.5ms between input and script trigger (which sets VPM switch state) on VPX side, then average of 0.5ms between switch state and VPM catching up, then again average of 0.5ms between VPM emulation and next VPX sync. So not that far from fast flips, but extended to all input/outputs between PinMame and emulation app (the real value is higher for the time being since VPX is still missing some threading improvment to move nearly all of the rendering outside of the main game logic thread).

### DIFF
--- a/src/cpuexec.c
+++ b/src/cpuexec.c
@@ -809,8 +809,38 @@ void cpunum_set_halt_line(int cpunum, int state)
  *
  *************************************/
 
+#if defined(_WIN32) || defined(_WIN64)
+// Sadly Windows does not offer a microsecond precise sleep function like unix does
+// using uSleep (from ticker.c) or Sleep results in bad precision and/or high CPU use
+// Taken from https://www.c-plusplus.net/forum/topic/109539/usleep-unter-windows
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+void usleep(unsigned int usec)
+{
+	HANDLE timer = CreateWaitableTimer(NULL, TRUE, NULL);
+	if (timer)
+	{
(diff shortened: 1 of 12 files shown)


toxieainc:
was out, so couldn't review/merge, great work!

PR #283, opened 2024-06-21
TimeFence: use semaphore instead of sleep

vbousquet:
This PR gives a cleaner implementation using semaphore allowing clean syncing. The drawback is that it contains quite a lot more modification since ti has to implement portable semaphore for PinMame. I have tested it under Windows and it works very well. For Apple & Unix device, only build have been tested since I don't have any device to run it.

### DIFF
--- a/src/cpuexec.c
+++ b/src/cpuexec.c
@@ -7,6 +7,7 @@
 ***************************************************************************/
 
 #include <math.h>
+#include <stdbool.h>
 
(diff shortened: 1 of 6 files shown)


toxieainc:
Looking good!

jsm174:
> For Apple & Unix device, only build have been tested since I don't have any device to run it.

standalone and 10.8.1 are all on the latest pinmame now, and I tested on macos and iOS, and both running great.

Issue #286, opened 2024-06-25
CSI table flippers unstable


toxieainc:
..since one of the recent changes the CSI table flippers are not held in the constant up position, but flicker
@vbousquet

vbousquet:
It was a known issue (it is stated in the file comments), with a known fix (previous implementation did hack flipper bats by counting writes and filtering accordingly), but I had it only applied for IJ. I have activated for CSI as well and it seems to work right now. I haven't merged the PR to let you review it before. Just tell me

toxieainc:
Thanks!
CSI and IJ are a bit special as we needed to hack around with the frequencies for these 2 machines.  :/
So that may explain that weird special behavior.

PR #287, opened 2024-06-26
Apply flipper bat timing hack to CSI (same as IJ4)

vbousquet:
Fix https://github.com/vpinball/pinmame/issues/286


toxieainc:
Thanks!

PR #291, opened 2024-07-29
Update bulb characteristic to new measures

vbousquet:
@toxieainc these are just better bulb modelling based on real life measurements (bought a bunch of bulbs and some VPW members helped with measurements) but things are so interdependent (for example with the flicker/fusion filter) that it would be nice if you could have a look if things looks ok for you.

I have tested this with a bunch of tables supporting PWM and it looked good to me (still fading, but better for some fast blinks that I think were smoothed out by previous constants).

### DIFF
--- a/src/wpc/bulb.c
+++ b/src/wpc/bulb.c
@@ -3,16 +3,41 @@
 #include <math.h>
 #include "bulb.h"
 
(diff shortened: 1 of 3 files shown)


toxieainc:
From quick tests, looking good, thanks!!

PR #302, opened 2024-08-23
nbaf: add full sim for NBA Fastbreak

tomlogic:
Looking for some feedback on my code before merging, and any ideas regarding how to get the game to fire the "pass" solenoids when I hit the flippers (shift keys).  If I used the `stateDef` structure it never recognized the passing.  When I integrated solenoid tests into the `handleBallState()` handler I could only trigger the passing by hitting the flipper button 3-4 times.

Notes from my last commit:
```
Initial implementation works well, other than passing between players
when "In The Paint" requires multiple shift key (flipper button) hits to
trigger the passing solenoids.  Simulates defender motor/positioning,
defender magnet used to hold the ball, diverters/gate, and the backbox
game.

For some reason, nbaf_stateDef wouldn't trigger state changes with the
passing solenoids, so I integrated those checks into
nbaf_handleBallState().
```


toxieainc:
@volkenborn Do you have an idea? I don't..

volkenborn:
Nope, sorry. This is way beyond me.

toxieainc:
@tomlogic Then just give us a hint when we should merge this.. Thanks!

tomlogic:
Doing a bit of cleanup before submitting.  I was able to use the `mech_XXX` APIs for the defender mech instead of my manual implementation.  I wonder if anyone has implemented having balls automatically drain if they're in a "Playfield" state and the flippers turn off?  I'm testing out the Championship Link feature, and the game kills the flippers at halftime and waits for balls to drain before going into the backbox game.  And again at the end of the fourth quarter, or when you compete for a ring.  It'd be nice to automatically clear the playfield in those moments.

I have not figured out the flipper button issue.  I attained a high score, and using the flipper buttons was also wonky to enter initials.  I am getting an "Unable to open the keyboard device. (error 32)" when PinMAME starts up, so perhaps that's related.

toxieainc:
I would love to help you with both issues, but i'm pretty much clueless in that area.

tomlogic:
@toxieainc, automatically draining when the flippers are dead (e.g., Tilt) seems to be a simple addition.  I'm going to move this from `prelim` to `full` as well, since there isn't anything left (that I'm aware of) to simulate.

tomlogic:
Ready to merge, but I'd love to have someone try it out on a VPin to make sure I didn't break anything there.  I'm also not sure how to update the failing checks.  They're looking for `nbaf.c` in `prelim` instead of `full`, but I can't find any documents in the repository that still have the old path.

toxieainc:
Looks good to me, we'll figure out why it fails then. Thanks a lot!!
Would be cool to see a video of how the virtual and real machine interact.  :)

PR #331, opened 2024-09-18
DMD: add some docs

vbousquet:


docs/dmd.md
jueank, line 1: I just arbitrarily read your docs in this commit, as I use to check regularly what is happening in VPX code.
Please, never leave our hobby. Never stop coding. And never leave our community. You are gold.

PR #336, opened 2024-10-02
fix string overlapping in debugger command handling

mattwalsh:
Without this change, at least with my current libc I will get a crash on the possibility that the strings overlap


toxieainc:
Oh wow, yup, thats indeed a bad bug, thanks!

PR #353, opened 2024-10-25
DMD: better backward compatibility for dmddevice.dll

### LINKED ISSUE #344: Use of GTS3 & WPC driver through dmdevice can not be correct for both DMD rendering / frame identification
Confirmed with Waterworld PuPPack.
PupCapture triggers barely work on the latest master. (VPinMAME-sc-3.6.0-1118)
It triggers without any problems in the pre-release version v3.6.0-998.

vbousquet:
I limited the change to what is described in the issue #344 (it fixes #344) but I think, there is some more room for cleanup since after reviewing the codebase, there seems to be some legacy left over for dmddevice.h/dmddevice.cpp

dmddevice.h/.cpp in ext/dmddevice contains the interface for dmd device drivers: some struct definitions and some function definitions

dmddevice.h/.cpp in PinMame source contains:
- a copy of `rgb24` and `layout_t` struct from ext/dmddevice/dmddevice.h (but not the rest of it)
- an helper to get a layout to render alphanumeric on a 4:1 DMD, the implementation to actually render is in usbalphanumeric.h inside ext/dmddevice folder imported inside PinMame as well as used by ext/dmddevice drivers
- a VPinMame only definition/implementation to exchange data through dmddevice.dll, only used inside VPinMame

To be more clean, I would propose to:
- avoid name conflicts, so rename PinMame dmddevice.h/.cpp to something else,
- keep the definition in only one place, so import `ext/dmddevice/dmddevice.h` instead of copy/pasting it
- split the alphanumeric layout/rendering from the other VPinMame only part (add it to core.h/.c),
- move the VPinMame only part to VPinMame folder in its own file,
- compile the PinMame part as C, and the VPinMame part as C++ (as a consequence of steps above).

To be even more clean, the device drivers could be in their own Github repo if they have one (since there are now quite a bunch of hardware drivers: PinDMD1, PinDMD2, PinDMD3, ZeDMD, RGB DMD, Pin2DMD, Freezy's DMD,...).

If you wish, I can do these.

### DIFF
--- a/docs/dmd.md
+++ b/docs/dmd.md
@@ -3,7 +3,7 @@
 Emulating the DMD pursues the following purposes:
 - providing emulators (PinMame, VPX,...) with informations allowing rendering of a virtual DMD as accurately as possible,
 - driving modern DMD hardware which implements their own shading capabilities,
-- providing frames that can be uniquely identified to perform colorization.
+- providing frames that can be uniquely identified to perform colorization or trigger events (f.e. PinUp display events).
 
 These aims lead the DMD emulation to evaluate 2 different things:
 - luminance frames, which are suited for accurate rendering, but may vary as the luminance model improves,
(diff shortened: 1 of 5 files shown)


PastorL69:
@vbousquet Just out of curiosity because we talked about this a month or so ago: You are right now reprogramming the entire way the DMD is being emulated correct? So the goal is to make everything work 1 to 1 compared to the real machines out there? If so, good stuff man :)

PastorL69:
Then I also imagine that when you are done doing this, the code will probably not be touched again for a long time.
Recently someone started doing his first Serum colorization, he chose Tee'd off (Gottlieb). 
From your recent pinmame versions, it seems that you have been correcting the emulating behaviour, and this has unfortunately caused his colorization to display incorrectly, since the source frames now often have different values. 

I know that this is important and that it needs to be done, but it always sucks when you start working on something and then stuff like this gets changed. It seriously adds a ton of extra work to the project.

toxieainc:
Yes, thats the downside. But Vince is doing a lot of groundbreaking work here, while being extremely cautious to be as backwards compatible as feasible. And so a lot of machines that had serious issues before now have much better DMD timings and support fading, etc, now.
So i fear we all still have to wait for some more weeks until the dust settles, and then hopefully we'll be good with groundbreaking changes for a while for DMDs, etc.

toxieainc:
@vbousquet Sounds all good, something that is also on my todo list since years, so please go ahead. As for outsourcing the device drivers: Where it makes sense (=still actively maintained) this should be better, yes.

vbousquet:
> Then I also imagine that when you are done doing this, the code will probably not be touched again for a long time. Recently someone started doing his first Serum colorization, he chose Tee'd off (Gottlieb). From your recent pinmame versions, it seems that you have been correcting the emulating behaviour, and this has unfortunately caused his colorization to display incorrectly, since the source frames now often have different values.
> 
> I know that this is important and that it needs to be done, but it always sucks when you start working on something and then stuff like this gets changed. It seriously adds a ton of extra work to the project.

The latest version with this changes will allow (mostly) full backward compatibility. Work should not be lost (or need very limited update regarding timing speeds). But this needs dmddevice.dll implementers (Freezy's DMDExt or lucky1) to update to the new API. Until then, GTS3 colorization will not work. Also, I don't know how colorization tools work but they will likely need a little update too for GTS3 games (to capture the right frame data). The core problem is that GTS3 DMD emulation has always been hacky: timings and shades were wrong, so fixing while maintaining backward compatibility was difficult but is now possible while needing every dmddevice to update to the new API.

Note that other hardware should be fully backward compatible without any update.
The only thing that may arise are very little fixes that should not break existing colorization (the little timings fxes for Data East 128x32 DMD, DE 128x16 and 128x64 double command processing, ...).

PR #392, opened 2024-12-21
Render dimmed DMD using shaded palette

vbousquet:


volkenborn:
@vbousquet Why did you remove that code again?
`if (keepColor) { // mini DMDs, don't change their colors`

Now the mini DMD colors for Simpsons Pinball Party are wrong again, and generally the dmd_perc33 and dmd_perc66 values don't show any effect in any DMD anymore. 🙁

volkenborn:
With this code, at least the mini DMDs can be adjusted in brightness:
`  else { // Only bitplane state: consider luminance equal to (scaled) bitplane output
    dmdDotRaw = &coreGlobals.dmdDotRaw[0];
    dmdDotLum = &coreGlobals.dmdDotLum[0];
    if ((core_gameData->gen & GEN_SAM) == 0) {
      const int shift = (core_gameData->gen & GEN_SPA) != 0 ? 4 : 6;
      const int perc33 = pmoptions.dmd_perc33 ? pmoptions.dmd_perc33 * 255 / 100 : 0;
      const int perc66 = pmoptions.dmd_perc66 ? pmoptions.dmd_perc66 * 255 / 100 : 0;
      const int step1 = perc33 ? perc33 / 5 : 0;
      const int step2 = perc66 > perc33 ? (perc66 - perc33) / 5 : 0;
      const int step3 = perc66 ? (255 - perc66) / 5 : 0;
      const UINT8 lum4[] = { 0, perc33, perc66, 255};
      const UINT8 lum16[] = { 0, step1, step1*2, step1*3, step1*4, perc33, perc33+step2, perc33+step2*2, perc33+step2*3, perc33+step2*4, perc66, perc66+step3, perc66+step3*2, perc66+step3*3, perc66+step3*4, 255 };
      const UINT8* lum = (core_gameData->gen & GEN_SPA) != 0 ? lum16 : lum4;
      for (int ii = 0; ii < layout->length * layout->start; ii++)
        dmdDotLum[ii] = lum[dmdDotRaw[ii]];
    }
`

The main game DMD has "full" DMD state. Haven't found the code where this is genereated yet...

vbousquet:
I changed it because it broke WPT mini DMD. I think the problem is that I did not fully refactored internal rendering when I added PWM support for DMD (and SAM mini DMD, but not for Whitestar mini DMD), therefore each time one of us fix a DMD, he may also break another one. I did not know there was an Red/Green mini DMD in Pinball Party, so the change I made broke it... The solution would be to keepcolor for SE miniDMD as they are not PWM shaded, and don't keepcolor for SAM miniDMD as they are shaded.

To move forward, after having a look, I dont think the RGB565 (as toxie tried for alphanum) is possible as most drivers rely on a paletted rendering. So, we are likely stucked to paletted rendering, keeping the base palette untouched to avoid breaking all drivers. Then for PWM shaded parts (alphanum, DMD, SAM miniDMD), we need to use one or more group of palette shades.

For the time being, we are using only one palette of 48 shades at the end of the main palette. This block does not apply dmd_perc33 / dmd_perc66 because shaded alphanumdid not applied these ( dmd_perc33 / dmd_perc66 are indeed applied for VPM). So either we keep only one 48 shade palette for Alpha/DMD and decide to apply dmd_perc33 / dmd_perc66 or not, or we go to create 2 palette of 48 shades, one for alphanum without dmd_perc33 / dmd_perc66 and one for DMD with dmd_perc33 / dmd_perc66. Finally, I think we need to have a look to the alpanum unshaded level (as alphanum shades starts at 0, so below unlit state)

vbousquet:
> The main game DMD has "full" DMD state. Haven't found the code where this is genereated yet...

The DMD state is generated through `core_dmd_update_pwm` like for all the DMD with PWM support

volkenborn:
Look at the Simpsons Pinball Party logo for example. The "Simpsons" text first appears brighter, then after half a second or so, it turns too dark. You can repeat the screen using the left / right flipper keys in attract mode to get there faster BTW.

simp0000

simp0001

vbousquet:
> Look at the Simpsons Pinball Party logo for example. The "Simpsons" text first appears brighter, then after half a second or so, it turns too dark. You can repeat the screen using the left / right flipper keys in attract mode to get there faster BTW.

I don't think so, this seems intentional from the game code. If you add `printf("DMD VBlank => %02x %02x\n", *(dmdlocals.RAMbankPtr + src + 0x005), *(dmdlocals.RAMbankPtr + src + 0x205));` to the DMD vblank in dedmd.c, it shows the data sequence the gamecode sends:
b6 b6 (previous screen: 2 + 1 brightness)
f0 b6 (fading between screen)
f0 00 (logo brightness : 2)
00 f0 (logo brightness : 1)

So it blinks the log then fade it down and show the 'Pinball Party' at full brightness

toxieainc:
For me, dimmed segment rendering is still broken, e.g.:
grafik

vbousquet:
What I propose is that I will do the following fix:
- implement a hack, just for Simpson's Pinball aventure: it is the only game with a multicolour LED matrix. To be fully clean, it would require 2 shaded frame per matrix (one for the green and one for the red) but it seems overkill as shading is not implemented for Whitestar,
- implement 2 shaded palette of 48 shades each, one for DMD using dmd_percXX, one for alphanum going from 20..100 (to account for unlit segments),
- redo the alphanum rendering using this fixed palette.
And hopefully, we should be good !

toxieainc:
Just for reference, this is with PinMAME32, there only the additional segments are weird:
grafik

vbousquet:
Just merged the fix

toxieainc:
Unfortunately, Hoops still looks exactly the same as the 2 screenshots above (VPM 'nothing' and PinMAME(32) with the additional segments 'nothing').

vbousquet:
Dimmed segment of GTS3's additional displays were not implemented. I just added them

PR #419, opened 2025-01-22
dw: use full name for game

tomlogic:
The game is called Doctor Who, abbreviated to DW, but not Dr. Who.

I've been spending time with the WPC-era games, and notice that most use the full name of the game.  The abbreviation to "Dr. Who" seems unnecessary.


toxieainc:
Oh, indeed, good catch..

PR #421, opened 2025-01-29
libpinmame: update to use combined CMakeLists and static/shared builds

jsm174:
This PR will convert libpinmame into a single unified CMakeFile.txt

In addition, there are the following changes:

- Each platform will now have the ability to generate shared and static libraries. 
- The test application is now built and included in the archives for all non mobile platforms 
- Unique name for 64 bit builds for windows `pinmame64.dll` / `pinmame.dll`
- Linux arm64 builds are now built in arm64 github runners 
- Fixed library filename suffixed for linux and iOS. (No more strange `-l"pinmame.3.6"`) 

I would prefer not to merge this until 3.6 is released. This should be consider a 3.7.0 (because of the versioning mentioned above)


jsm174:
Now that 3.7.0 has started, I will merge this. I also updated the `VERSION_START_SHA`, so that this build would be 2.

PR #431, opened 2025-03-02
Add new COM Version method including build number as double

JockeJarre:
This adds a PMBuildVersion COM method to VPM similar to VPBuildVersion and B2SBuildVersion which includes the build number as decimal.

-> 3700,0012

To use this version number in Tables, there will be a method in B2S forwarding this value.


toxieainc:
Thanks!

PR #433, opened 2025-03-03
Add new COM Version method including build number as double **new format**

JockeJarre:
Changed output format 3700.4711 -> 30700.4711 leaving for more minor version numbers and keeping the output similar to Visual Pinball
Added documentation to the html file and to readme.txt


JockeJarre:
We now have the same kind of BuildVersion method in PM, VP and B2S with similar output:

print VPBuildVersion
10801,2965
print Controller.B2SBuildVersion
20104,0332
print Controller.PMBuildVersion
30700,0013

toxieainc:
Thanks!

PR #434, opened 2025-03-04
Easier to upgrade: Rename VPMAlias.txt

JockeJarre:
I have now moved all aliases from Alias.cpp to VPMAlias.txt -> only one file to edit.
This file is now automatically integrated into Alias.cpp before compilation in the workflow. -> only one file to edit.

On assembly VPMAlias.txt is amended as VPMAlias_template.txt

All this means that VPMAlias.txt can start empty and end users only need to add personal entries.
Each update will become easier, since all files can be unzipped into the folder.


JockeJarre:
@jsm174, is there any special handling in standalone for Aliases? I have not found anything in this repository.

toxieainc:
Neat, thanks!!

jsm174:
> @jsm174, is there any special handling in standalone for Aliases? I have not found anything in this repository.

@JockeJarre - correct, libpinmame doesn't support aliases. I've always wanted to add it, but have not got around to it.

PR #463, opened 2025-07-16
Clean up some game names

tomlogic:
Just some minor changes to update game names so they match what IPDB and OPDB use.  Verified by checking backglass and/or manuals.


toxieainc:
Great, thanks!!

PR #470, opened 2025-07-28
startup: exit if ROM files are missing

tomlogic:
This line was commented out 20 years ago in 4ec6b7c.

I'm not sure of the reasons behind the change, but I'm doing some automation with PinMAME and wanted to return to the previous behavior of automatically exiting for ROMs that don't exist.

Is there a reason to keep this line commented out?  Do users launch from batch files in Windows Explorer and therefore miss seeing error messages if there isn't a "Press any key to continue" prompt?  there are other instances of MAME code setting that flag (see `bail_and_print()` in mame.c), so maybe we don't need special treatment for missing ROMs?

Alternatively, I was looking for a way to set `options.gui_host` from the command line, but see that's only present in `Win32ui.c` and `VPinMAMEConfig.cpp`.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -1226,7 +1226,7 @@ static int display_rom_load_results(struct rom_load_data *romdata)
 		if (romdata->errors)
 		{
 			strcat(romdata->errorbuf, "ERROR: required files are missing, the game cannot be run.");
-			// bailing = 1;
+			bailing = 1;
 		}
 		else
 			strcat(romdata->errorbuf, "WARNING: the game might not run correctly.");


tomlogic:
And I guess I'll throw in another change (6f51f3b) related to starting up.  If you're launching PinMAME with the `-nosound` option, it shouldn't report warnings for games with imperfect or missing sound.

toxieainc:
Looks to me like the commit back then tried to find out if its 'just' sound roms that are faulty and then handle that via disabling sound.
@volkenborn What do you think what to do with this behavior?

tomlogic:
I have two changes in separate commits.

The first removes the "Press any key to continue..." message after PinMAME reports that ROMs are missing.  In that case, PinMAME can't run at all, and the only reason to include a prompt like that is if someone launched PinMAME outside of a cmd.exe terminal and will miss the error message otherwise.  In cases like that. they're probably launching it via a `.cmd` or `.bat` file, and can add a pause there.

For the second, I'm passing the `-nosound` option to PinMAME.  I originally did that to quiet warnings from PinMAME about issues with the sound ROMs.  But when I ran `jokrz_l6`, it was presenting a warning that "sound emulation isn't 100% accurate" that required typing "OK" to continue.  My change is that it shouldn't display warnings about sound emulation when sound is explicitly disabled.

Your example of "Sir Lancelot" should be unaffected by my changes.

volkenborn:
> @volkenborn Would it be okay to merge then?

If it works, sure. 👍

PR #541, opened 2026-01-10
Rename lum function in DmdDevice API.

freezy:
In general, the problem with changing signature and data format of an API used by third parties is that the third parties break. However, in this case the problem is that the behavior of PinMAME changes when a third party with this feature is loaded - frames will only be emitted through the new format.

However, the new format isn't defined; For PinMAME 3.6, it's gray8  frames, and for 3.7+ the data is sent in (32-bit) floats. There is no way for third parties to know what they'll be getting. That means a third party that implements `Render_Lum_And_Raw` as gray8 will break with PinMAME 3.7 and a third party that implements `Render_Lum_And_Raw` as floats breaks for PinMAME 3.6.

By renaming the method of the API, the data format (and signature) is clearly defined and allows third parties to release implementations that work for both PinMAME 3.6 (and older) and 3.7+.


vbousquet (commented):
Looks clean to me

PR #562, opened 2026-04-05
fixed flickering of segment displays rendered as synthetic DMD

mkalkbrenner:
I didn't update pinmame since August 2025 in PPUC. Now I did it and adopted the libpinmame changes. But I noticed that Williams Sys3 and Sys6 games are now flickering, like a permanent on/off per segment.
I added some debug code in PPUC and saw that the top right corner that should show a stable `0`in attract mode toggles (cornerChecksum):
```
OnDisplayUpdated(): index=5, type=14, top=0, left=0, width=128, height=32, depth=2, length=0
SyntheticDMD: index=5 blank=0 lit=59 min=0 max=128 checksum=a2ca78d9 cornerLit=0 cornerChecksum=4a2bb865 hw=20000
OnDisplayUpdated(): index=5, type=14, top=0, left=0, width=128, height=32, depth=2, length=0
SyntheticDMD: index=5 blank=0 lit=687 min=0 max=3 checksum=f435a51a cornerLit=57 cornerChecksum=ed057750 hw=20000
OnDisplayUpdated(): index=5, type=14, top=0, left=0, width=128, height=32, depth=2, length=0
SyntheticDMD: index=5 blank=0 lit=59 min=0 max=128 checksum=a2ca78d9 cornerLit=0 cornerChecksum=4a2bb865 hw=20000
OnDisplayUpdated(): index=5, type=14, top=0, left=0, width=128, height=32, depth=2, length=0
SyntheticDMD: index=5 blank=0 lit=687 min=0 max=3 checksum=f435a51a cornerLit=57 cornerChecksum=ed057750 hw=20000
```

For segment-only games, libpinmame created an extra synthetic 128x32 DMD but stored it in the same display slot as the last segment display.

So updates alternated between:
- real synthetic DMD frame data
- segment-display data misread as DMD data

That caused the visible flicker. I gave the synthetic DMD its own dedicated display index and now it is working for PPUC. But you have to judge if that is the correct fix.

Using this fix, I also get way less display updates which is very important.

### DIFF
--- a/src/libpinmame/libpinmame.cpp
+++ b/src/libpinmame/libpinmame.cpp
@@ -533,11 +533,10 @@ extern "C" void OnStateChange(const int state)
 			if (displayCount <= layout->index)
 				displayCount = layout->index + 1;
 		}
-		const int syntheticDmdIndex = displayCount;
-		if (hasDMDOrVideo)
-			displayCount++;
-		else
-			displayCount = syntheticDmdIndex + 1;
+		// Reserve one extra slot unconditionally.
+		// Using hasDMDOrVideo as the gate is wrong here: segment-only games also create an extra
+		// synthetic 128x32 DMD below, so they need one additional display slot as well.
+		displayCount++;
 		_displays.resize(displayCount);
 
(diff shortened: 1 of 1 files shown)


mkalkbrenner:
The change is immediately visible in the DMD and CPU usage drops.

PR #573, opened 2026-05-12
Data East 128x16 DMD shade correction.

PastorL69:
I've done research into this and came to the realization PinMAME incorrectly handles 128x16 shades. 
Looking at the code, it's odd that DE x16 does have its own case called SUM_1_2, though this isn't being correctly followed up.

### DIFF
--- a/src/wpc/core.c
+++ b/src/wpc/core.c
@@ -3730,8 +3730,8 @@ UINT8* core_dmd_update_identify(const core_tLCDLayout* layout, unsigned int * ra
         frame1 = dmd_state->rawFrames + ((dmd_state->nextFrame + (dmd_state->nFrames - 1)) % dmd_state->nFrames) * dmd_state->rawFrameSize;
       }
       else { //if (dmd_state->raw_combiner == CORE_DMD_PWM_COMBINER_SUM_1_2) { // double length frame are the 2 last ones, single length frame is the one before (Data East 128x16)
-        frame0 = dmd_state->rawFrames + ((dmd_state->nextFrame + (dmd_state->nFrames - 2)) % dmd_state->nFrames) * dmd_state->rawFrameSize;
-        frame1 = dmd_state->rawFrames + ((dmd_state->nextFrame + (dmd_state->nFrames - 1)) % dmd_state->nFrames) * dmd_state->rawFrameSize;
+        frame0 = dmd_state->rawFrames + ((dmd_state->nextFrame + (dmd_state->nFrames - 1)) % dmd_state->nFrames) * dmd_state->rawFrameSize;
+        frame1 = dmd_state->rawFrames + ((dmd_state->nextFrame + (dmd_state->nFrames - 3)) % dmd_state->nFrames) * dmd_state->rawFrameSize;
       }
       for (int kk = 0; kk < dmd_state->rawFrameSize; kk++) {
         unsigned int intens1;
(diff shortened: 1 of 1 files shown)


vbousquet:
Hello and sorry for being a bit long on this. Just had a look and I don't really get the intent of the PR.
What we do for the time being is support to combination modes, designed to support backward compatibility (not computing brightness but doing the same as previous version to not break existing colorizations):
- CORE_DMD_PWM_COMBINER_SUM_2_1 will have 2 identical frames then a different one, so we double frame `#0` and add frame `#2`
- CORE_DMD_PWM_COMBINER_SUM_1_2 will have 1 frame then 2 identical frames, so we double frame `#1` and add frame `#0`

(we could do it more simply adding the 3 frames in all  situations, but it is supposed to be a bit better performance wise like this)

This is done by adjusting the pointers frame0 / frame1. The PR revert this behavior, in turn ignoring the combiner order (1/2 vs 2/1) and therefore I would say breaking it.

So, I guess we should get back to the initial issue you are stating which I understand to be DataEast 128x16 shades:
- regarding backward compatibility, does current implementation works with existing colorization or not (I remember doing some testing on this while I implemented it, but maybe I missed somehting). If it is right, then nothing should be changed on the combiner side.
- regarding brightness computation, does the current implementation results in correct brightness or not (I also remember doing some testing on this while I implemented it, but maybe I missed somehting). If the brightness is reversed, then we should swap the long and short frame, *and* change the combiner from 1_2 to 2_1 (and adapt comments accordingly)

toxieainc:
I just compared the VPM output (via setup->test) with a Hook youtube vid, and to me the shades look correct.

PastorL69:
Hey, sorry for also not following up on this. Most of the conversation was held with Vincent in private messages. I will share my findings here. I'd like to believe pinmame handles it incorrectly. 

# The analysis
First off, I have been developing and collecting real pinball DMD signals together with @mkalkbrenner over at the dmdreader repository: https://github.com/PPUC/dmdreader 
in the logicanalyzer folder you will find the hook signals: https://github.com/PPUC/dmdreader/tree/ppuc/logicanalyzer/Sega%20%26%20Data%20East

So, the entire analysis is based on the real game signals and the way it decodes. As you know, with unstable row type systems, there is an LSB and MSB row.

For x16, it just so happens that whenever a dim pixel is lit, it always happens to be the data originated from the LSB row, and not the MSB row. Further proving the 33% value being correct. This is consistent across multiple samples I've had a look at in the Hook signals. Check out the screenshots. 

image

image

image

Through a friend I checked what type of dumps a PIN2DMD real pinball display produces. There it also is a dim pixel having value "1" out of 0, 1, 2 and 3. Not matching the PinMAME produced frames, as PinMAME would put a "2" instead of "1'".

Across the years I've heard people saying that the pin2dmd created colorizations never worked well for the x16 games, certainly this is why, a greyscale mismatch.

PastorL69:
Aside from that research, the PR I created is likely not something that can be merged. But with this pr I was able to dump frames and they now compared to the way I have it configured over at dmdreader and the pin2dmd real pinball display.

However, there is always the possibility I am wrong. If there is anything in software you guys see to prove the opposite then I'd like to know more as I will have to correct it in dmdreader.

toxieainc:
Thanks!

As i'm no low level expert on this, i'll trust with whatever you can come up with.
So i'm mainly concerned about correctness (i.e. for preservation), and then we'll also have to have the backwards compatible mode working same as 'before' to not break existing colorizations.

PastorL69:
As written to @vbousquet privately, I am closing the pr after also concluding that PinMAME does handle it correctly. 

https://www.youtube.com/watch?v=k3qEFzPFfhQ&t=289s

After also looking at one of these displays in person, to me it was now also obvious that shade 0, 2 and 3 are the dominant ones. So not 0, 1 and 3 as I had claimed. The signals in the way they are handled do prove to be 0, 1 and 3. But if there's anything I learned also implementing the signals of this display on a microcontroller, it's that I know almost everything was different from the regular 128x32 DMD signal implementation. So leaving it at that :)

toxieainc:
Nevertheless, thanks! Its always good to doublecheck or proof things! Anything we could add to the source as documentation from your learnings?

PR #582, opened 2026-06-07
Improve headless support: fix zero-color-mask palette hang + add -skip_gamewarnings

gerwout:
Makes the SDL/Unix build usable headless (no interactive display). Headless runs previously hung at two unrelated points; this branch fixes both.

  ### 1. Fix infinite loop in palette setup on a zero color mask

  `sysdep_palette_make_pen_from_info()` (`src/unix/sysdep/sysdep_palette.c`) derives each RGB component's shift by starting at the top bit (`1 << 31`) and right-shifting until it meets a set bit of the color mask. With a signed `int pen`
  that right shift is **arithmetic** and sign-extends, so `pen` never reaches 0. If a color mask is `0` — e.g. an 8bpp/indexed display where the SDL driver leaves the RGB masks at 0, or any zero-mask offscreen target — the loop never
  terminates and the program hangs during display-palette setup.

  Making `pen` **unsigned** makes the shift logical, so the loop always terminates. Real displays have non-zero masks and are unaffected; this only fixes the zero-mask / headless case.

  ### 2. Add `-skip_gamewarnings` for headless runs of not-yet-working games

  The startup game-warnings screen (`showgamewarnings`) is shown for any driver flagged `GAME_NOT_WORKING`, `GAME_WRONG_COLORS`, `GAME_IMPERFECT_*`, etc. It loops on `update_video_and_audio()` waiting for the user to type "OK", which
  never arrives in a headless run, so the emulator spins forever before reaching the game. The existing `-skip_disclaimer` and `-skip_gameinfo` switches cover the other two startup screens, but the warnings screen had no equivalent.

  `-skip_gamewarnings` is wired exactly like those two: a field in `struct GameOptions` (`src/mame.h`), an `rc_bool` registration (`src/unix/config.c`), and a guard on the `showgamewarnings()` call in `run_machine_core` (`src/mame.c`).

  This **adds headless support for games that do not (fully) work yet**: drivers still flagged `GAME_NOT_WORKING` can now be booted into their emulation loop headlessly (e.g. for automated frame captures or batch validation). Working
  games never trigger the warnings screen and are unaffected.

  ### Headless usage

  ```bash
  # sdl12-compat forces X11 and won't use an 8bpp visual, so use a 24bpp virtual display
  Xvfb :99 -screen 0 1280x1024x24 & export DISPLAY=:99
  SDL_AUDIODRIVER=dummy ./xpinmame.sdl <game> --rompath <roms> \
    -nosound -skip_disclaimer -skip_gameinfo -skip_gamewarnings -frames_to_run 200
  ```


toxieainc:
Hi! I would guess that you need to wire the new option everywhere, not just in the linux config?

gerwout:
> Hi! I would guess that you need to wire the new option everywhere, not just in the linux config?

That is a fair point actually, ok I will do some work, add some support for Windows, test it and update my PR.

gerwout:
@toxieainc I have also added support for the Windows cli and Windows gui version for the new skip game warnings functionality. So, it should work for all scenario's now.

PR #584, opened 2026-06-13
Bitreverse updates

Disservin:
- ever since compilers have __builtin_bitreverse functions they automatically use rbit for arm when it's available, so move the builtin code first 
- newer gcc versions also have __builtin_bitreverse, so use a __has_builtin check instead


toxieainc:
Thanks!

PR #585, opened 2026-06-13
Feature/wpc updates

tomlogic:
A few WPC-related updates I've made recently.

- Using a macro for repeated blocks of sound ROM entries eliminates some copypasta and makes the file easier to read.
- I recently ran across a Road Show LA-1 ROM, so I've added an entry for that.  I'll release a ZIP file for the ROM to the usual places.
- I'm doing some automated testing and getting hung up on No Fear launching balls during its startup sequence, so I ported code from the NBA Fastbreak sim that automatically drains balls on the playfield when the flippers are turned off.

That auto-drain code could go into other sims so it isn't necessary to manually drain balls on tilt.  It shouldn't negatively impact games with video mode since those balls will be held somewhere and not in `stFree`.


toxieainc:
Thanks!

PR #586, opened 2026-06-13
Unix: enlarge sound FIFO to fix garbled audio on Linux (ALSA/PipeWire)

gerwout:
The software sound FIFO between the emulator and the audio DSP (sound_stream) was created with only 3 per-frame buffers (~50ms). With the free-space-paced DSPs (ALSA/OSS/esound) on modern Linux -- in particular PipeWire's ALSA-compat scheduling -- this FIFO runs dry on a large fraction of update cycles, and sound_stream_update() then replays the last buffer ("looping sample") to fill the device's free space. Those repeated samples garble the audio, most audibly on speech.

Increase the FIFO depth to 32 per-frame buffers (overridable via the PINMAME_SND_BUFCOUNT environment variable). The FIFO stays near-empty in steady-state flow-through, so latency is essentially unchanged, but it absorbs the per-frame bursts and small generation/consumption jitter. On Sleic Bike Race (yes, I am working to get the Sleic games running as well...) the underrun-looping rate drops from ~62% to ~0.6% (startup only) and the audio plays cleanly.

### DIFF
--- a/src/unix/sound.c
+++ b/src/unix/sound.c
@@ -24,6 +24,33 @@ static struct sysdep_mixer_struct *sound_mixer = NULL;
 static int sound_samples_per_frame = 0;
 static int type;
 
+/* Software FIFO depth (number of per-frame buffers) between the emulator and
+   the audio DSP (see sound_stream.c). The historic value of 3 (~50ms) is far
+   too small for modern Linux audio: with the free-space-paced DSPs (ALSA/OSS/
+   esound), notably PipeWire's ALSA-compat scheduling, the FIFO runs dry on most
+   update cycles and sound_stream_update() replays the last buffer ("looping
+   sample") to fill the device's free space, garbling the audio (most audibly on
+   speech).
+
+   Measured on Bike Race (PipeWire, 48kHz): underrun looping collapses by a
+   depth of 8, and the FIFO occupancy plateaus at an ~8-buffer peak regardless
+   of the cap, so larger values add headroom but not latency (total latency is
+   dominated by the fixed ~265ms device buffer). 16 sits comfortably above the
+   peak demand with no audible latency cost and no CPU/memory cost.
+
+   Override with the PINMAME_SND_BUFCOUNT environment variable. */
+#define SND_DEFAULT_BUFCOUNT 16
+static int sound_stream_bufcount(void)
+{
+   static int v = 0;
+   if (!v) {
+      const char *e = getenv("PINMAME_SND_BUFCOUNT");
+      v = e ? atoi(e) : SND_DEFAULT_BUFCOUNT;  /* default when unset  */
+      if (v < 2) v = 2;                        /* structural minimum  */
+   }
+   return v;
+}
+
 static int sound_set_options(struct rc_option *option, const char *arg,
    int priority)
 {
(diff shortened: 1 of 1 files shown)


toxieainc:
I'm no expert on the unix code of PinMAME, but isn't 32 a bit excessive? Doesn't this imply that the sound latency is increased by a lot? Or am i missing something?

Also, shouldn't the check `if (v < 2) v = 32;` be rather something like `if (v < 32) v = 32;` ?

gerwout:
@toxieainc 

Thanks for the review,  you're right that 32 was unnecessarily large, so I've dropped the default to 16; and you're right the conditional was off, though the fix turned out a little different from your suggestion. Details below.

  Is 32 excessive / does it add a lot of latency?

I instrumented the software-FIFO occupancy and the device-side latency (snd_pcm_delay) and swept buf_count on Bike Race (PipeWire, 48 kHz; 800 samples/frame ≈ 16.7 ms per buffer). "underrun-loops" = the fraction of update cycles where the FIFO ran dry and sound_stream_update() replayed the last buffer, i.e. the repeated samples that cause the garble.
```
 buf_count     underrun-loops  FIFO latency    total latency*
  3 (old)           61.7 %                   ~47 ms             ~312 ms
  4                    42.2 %                   ~57 ms             ~322 ms
  6                      3.2 %                   ~67 ms             ~332 ms
  8                    ~0 %                      ~74–96 ms       ~340–360 ms
  16                  ~0 %                      ~96 ms             ~362 ms
  32                  ~0 %                      ~90 ms             ~356 ms
```
* total = FIFO + the fixed ~265 ms device buffer (ALSA 250 ms + PipeWire), which buf_count does not change. CPU (~30.6 s) and RSS (~159 MB) were flat across every value — no CPU/memory/IO cost.

The counter intuitive bit: FIFO latency does not scale with the cap. Steady-state occupancy plateaus at an ~8-buffer peak (it's set by the producer/consumer phase, not by the buffer size), so 32 has basically the same FIFO latency (~90 ms) as 8 (~74 ms). So buf_count is really a cleanliness/headroom knob, not a latency knob — across 8…32 it moves total latency by <20 ms (~5 %), dwarfed by the fixed ~265 ms device buffer.

So 32 wasn't actually a latency problem, but you're right it's more headroom than needed. The cleanliness knee is at 8 (looping collapses 42 % → 3 % → ~0; the residual is startup-only and decays to ~0 over a session), and the FIFO peaks at ~8 buffers, so I set the default to 16 — comfortably above the peak demand with no audible latency cost. (If anyone does want to cut latency, the lever is that ~265 ms device buffer via -bs/-abuf, which is independent of this change.)

  The conditional

Good catch, if (v < 2) v = 32; was confusing because that 32 was doubling as the default. Your if (v < 32) v = 32; fixes the readability but would clamp the minimum to 32, which stops anyone tuning it down — the opposite of the latency concern. So I split the two concepts:

  #define SND_DEFAULT_BUFCOUNT 16
  ...
  v = e ? atoi(e) : SND_DEFAULT_BUFCOUNT;  /* default when unset */
  if (v < 2) v = 2;                        /* structural minimum */
  
  I also added 2 sound samples, so that the difference (before/after) can be reviewed
  [bikerace_gameplay_audio.wav](https://github.com/user-attachments/files/28928974/bikerace_gameplay_audio.wav)
[bikerace_ALSA_devicepath_capture_trimmed.wav](https://github.com/user-attachments/files/28928976/bikerace_ALSA_devicepath_capture_trimmed.wav)

toxieainc:
Great, thanks!

PR #591, opened 2026-06-30
wpc/corv: add Corvette LA-2

tomlogic:
Please credit to Tom Collins/Scorbit.  We're finding older ROMs in the wild that aren't listed in PinMAME.


toxieainc:
Great, thanks!

PR #593, opened 2026-07-03
Fixed libpinmame display update callbacks for non-DMD segment displays

mkalkbrenner:
## Summary

Fixes `libpinmame` display update callbacks for non-DMD segment displays.

Old alphanumeric/7-segment games such as Williams Time Warp expose real segment displays through `core_dispLayout`, while PinMAME may also generate a DMD view from the same data. The generated DMD updated correctly, but clients using the native segment display callbacks received no segment data.

## Issue

`libpinmame_update_display()` handled changed DMD/video display data by copying `p_data` into `pDisplay->pData` and passing that buffer to `cb_OnDisplayUpdated`.

For non-DMD segment displays, the function always called:

```cpp
cb_OnDisplayUpdated(index, nullptr, &pDisplay->layout, userData)
```


vbousquet (commented):
I'm not really easy with this code that reimplement the legacy VPinMAME as a library.

I would say that the real fix here would be to **not** implement the alpaha to DMD rendering that exists inside VPinMAME. This rendering is incomplete and somewhat outside the purpose of PinMAME so it would be cleaner to just drop it instead of hacking it like this.

@jsm174 What do you think ? I will let it up to you to merge this PR which solves the bug by persisting the hacky design, or to remove the alpha to DMD feature (which I don't thihnk has any user atm). If you're ok with the removal but don't have the time, just ask.

jsm174:
For completeness, I merged, but after the plugin states get merged, maybe we go to a version 2.0 of libpinmame and do everything correctly.

PR #601, opened 2026-07-14
AT91 JIT: fix crash on iOS (exec-permission probe)

jsm174:
Also fix iOS debug build link errors: 

`static const uint32_t AJ_SLOT_BASE    = 2;` vs `static constexpr uint32_t AJ_SLOT_BASE    = 2;`


toxieainc:
Late to the party, but thanks!!

PR #614, opened 2026-08-21
i8085: fix RIM/SIM, the EI delay, TRAP re-arming and callback loss

gerwout:
Please note @volkenborn looked at this code already, he is aware that I am going to do separate PR's to get a currently not working game (Recreativos Franco - Super Star) working. 

```markdown
Five defects in the 8085A core, all found while bringing up a driver that drives
its display and reads its switches through the serial pins. None of the fixes is
specific to that machine.

- **RIM returned `I.IM` raw.** That register holds the masks and IE in the
  positions the RIM byte expects, but `IM_TRAP` (0x10) sat where the RST5.5
  pending flag belongs and `IM_SOD` (0x40) where RST7.5 pending belongs, and SID
  was only ever reported if a driver had pushed a value in beforehand. The byte
  is now composed properly.
- **SIM ignored bit 4 (R7.5).** RST7.5 is edge triggered and latches, so unlike
  RST5.5/6.5 it is never cleared by the line going low — only by SIM or by the
  interrupt being taken. Without handling bit 4 the latch sticks and the handler
  re-enters forever.
- **`i8085_reset()` wiped the callbacks.** It memsets the whole register struct,
  which dropped the irq, sod and sid callbacks a driver installs at init. Those
  are wiring, not CPU state.
- **EI took effect immediately.** On real hardware an interrupt is not
  recognised until the instruction after EI has completed, and service routines
  rely on it: an ISR ending `EI / POP PSW / RET` needs the POP to run first or
  the frame is left half unwound.
- **TRAP could only ever fire once.** `i8085_set_TRAP()` gated on `I.ISRV`, an
  "in service" lock that only EI clears — but a TRAP handler is free to run with
  interrupts disabled and simply RET. The lock then stayed set forever, blocking
  not just further TRAPs but RST5.5/6.5/7.5 as well. It now gates on the pending
  flag, which `Interrupt()` clears as it takes the interrupt, so a new edge
  re-arms while a second edge arriving before the first is taken is still
  ignored.

Adds `i8085_set_SID_callback()` so hardware that clocks serial data in is
sampled at the moment RIM executes rather than ahead of time.
`i8085_set_SID()` still works for drivers that prefer to push.

### Risk to existing drivers

This core is not unused, and these changes are not free of consequence for the
two drivers already on it:

- `src/wpc/regama.c` (Regama Trebol, added 2020) instantiates an `8085A` and
  drives **both** serial pins — `i8085_set_SID()` for the battery-good line and
  `i8085_set_SOD_callback()`.
- `src/wpc/micropin.c` has an `8085A` machine.

The RIM composition change, the EI delay, the TRAP re-arm and the callback
preservation all apply to them. I have no ROMs for either and could not test
them; a smoke test from someone who has would be worth having before this is
merged.

### DIFF
--- a/src/cpu/i8085/i8085.c
+++ b/src/cpu/i8085/i8085.c
@@ -211,6 +211,8 @@ typedef struct {
 	INT8	filler; /* align on dword boundary */
 	int 	(*irq_callback)(int);
 	void	(*sod_callback)(int state);
+	int 	(*sid_callback)(void);
+	UINT8	after_ei;	/* EI takes effect one instruction late - see below */
 }	i8085_Regs;
 
 int i8085_ICount = 0;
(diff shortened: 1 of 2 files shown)


toxieainc:
@volkenborn Are you okay with this as-is? Thanks @gerwout !!!

volkenborn:
> @volkenborn Are you okay with this as-is? Thanks @gerwout !!!

Yes I am. 👍

PR #615, opened 2026-08-21
rfranco: new driver for Recreativos Franco Super Star (1986)

gerwout:
```markdown
### Important note -> build will currently fail
**This needs the 8085A core fixes in #<PR 614> and will not compile without
them:** the driver installs a SID callback via `i8085_set_SID_callback()`, and
at run time it depends on TRAP re-arming and on the one-instruction EI delay.
Did decide to push the PR, so it can be reviewed. @volkenborn did partially review this code already.


A new driver for Recreativos Franco's *Super Star* (Spain, 1986), with two ROM
sets — `supstarf` (rev. 1, the revision the factory manual documents) and
`supstarfa` (rev. 2, newer firmware). Both rom sets are attached to this PR.


### The hardware

8085A main CPU with a 27128 at IC19 and 5517 battery-backed RAM at IC11, plus an
8035 sound CPU with a 2532 at IC4 driving two AY-3-8910s. Four 8212s carry the
command path between the two processors. The I/O is serial throughout: the
display chain hangs off SOD, the playfield switches are clocked back in through
SID, and any OUT instruction is the shared shift clock.

Two details worth knowing:

- TRAP carries the mains zero cross and is load bearing at boot. The reset path
  tests an NVRAM magic byte and RSTs back to `0000` when it fails, and it is the
  TRAP handler that seeds the magic — so without TRAP re-arming the machine
  never comes up at all.
- The 2532's data pins are wired to the 8035's AD0–AD7 in reverse order, so the
  sound image is bit-reversed at init. Straight off the chip it has 5 RET
  opcodes in 2K and no jump at either vector; reversed it has 172 JMP, 88 CALL,
  28 RET and the expected vector layout.

### State

Both sets are playable. A coin gives a credit, the start button serves a ball,
playfield contacts score, the drop-target banks light their specials, collecting
one awards a replay, each ball ends into the next with its bonus paid, and the
last ball ends the game with the final score held. Lamps, coils and the two
mains phases follow the ROM's own tables, and all four operator door modes work.

The two operator door switches are modelled as toggles in a pseudo coin-door
column (switches 1 and 2, keys `7` and `8`), following Williams System 4–11 —
they have to be movable while the machine runs, because the adjustments menu
re-reads them on every pass to decide what the start button does.
[supstarfa.zip](https://github.com/user-attachments/files/31323737/supstarfa.zip)
[supstarf.zip](https://github.com/user-attachments/files/31323738/supstarf.zip)

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Nuova Bell, Inder, Juegos Populares, LTD, Peyper, Sonic, Allied Leisure, Fascina
 Sleic, Playmatic, NSM, Grand Products, Jac van Ham, Videodens, Astro, Micropin,
 Christian Tabart, Jeutel, Valley Manufacturing, MAC / CICPlay, Stargame, Barni,
 Seeben/Sirmo, Splin Bingo, Playbar, Cirsa, Nondum / CIFA, Maibesa, ManilaMatic, Joctronic, Mirco,
-Sport Matic, Regama, Illinois Pinball.
+Sport Matic, Regama, Recreativos Franco, Illinois Pinball.
 
 *Note: Emulation is not 100% working and correct for all hardware, but very close for the vast majority.*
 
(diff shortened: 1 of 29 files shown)


toxieainc:
@volkenborn Are you okay with this as-is? Thanks @gerwout !!!

volkenborn:
> @volkenborn Are you okay with this as-is? Thanks @gerwout !!!

Yes, very much so! The change for the game setup switches from DIPs to toggles was one I requested. 😉

PR #616, opened 2026-08-21
remote_debug: filter by CPU, expose the mechanics flag, fix column 0 names

gerwout:
``markdown
Three independent improvements to the remote debug API. No emulation behaviour
changes.

- **Per-CPU filtering for breakpoints, instrumentation points and tracepoints.**
  The per-instruction hook read `activecpu_get_reg(REG_PC)` with no idea which
  processor was running, so on a multi-CPU board any point below the smaller
  CPU's ROM size counted hits from both. On the board this was found on, that
  made every measurement of the main CPU's low addresses meaningless — `0x0286`
  is an error path in the game ROM and a `MOV` inside a live voice routine in
  the sound ROM. All three point types now carry an optional `cpu` field,
  exposed as `&cpu=N` on the endpoints and reported back in the JSON. Omitted,
  it keeps the old any-CPU behaviour.
- **`/api/mechanics`** reads and sets `g_fHandleMechanics`. The standalone build
  pins the flag at `0xff` with no route to change it, which leaves the
  front-end-owned ball path untestable in any driver that has one.
- **`switch_name()` no longer mislabels column 0.** It gave switches 1–8 the WPC
  coin-door names ("Coin 1".."Escape") whatever the game. Column 0 is the coin
  door on WPC and whatever its driver makes it elsewhere, so the table is now
  gated on `GEN_ALLWPC` and otherwise falls through to the empty name that the
  rest of such a driver's switches already report.
  
  Created as part of PR 614 and PR 615 (Add Super Star) to help with debugging, should be possible to independently merge.

### DIFF
--- a/src/remote_debug/api_handler.c
+++ b/src/remote_debug/api_handler.c
@@ -539,22 +539,24 @@ static void handle_api_debugger_trace(const http_request_t *req, http_response_t
 
 static void handle_api_debugger_breakpoints(const http_request_t *req, http_response_t *resp)
 {
-	char cmd_buf[32], addr_buf[32], bank_buf[32], cond_buf[32], ignore_buf[32];
+	char cmd_buf[32], addr_buf[32], bank_buf[32], cond_buf[32], ignore_buf[32], cpu_buf[32];
 	get_query_param(req->query, "cmd", cmd_buf, (int)sizeof(cmd_buf));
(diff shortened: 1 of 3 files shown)


toxieainc:
Looks good to me, @whaslbeck for review, please..

whaslbeck:
looks good to me

toxieainc:
Thanks!

PR #626, opened 2026-08-31
Accept CRLF line endings in alias files

francisdb:
Both alias readers (`CheckGameAlias` in libpinmame for `alias.txt`, `checkGameAlias` in VPinMAME for `VPMAlias.txt`) terminate the real rom name with `strtok` on `" ,\n#;'"`, which does not include `\r`.

On Windows `fopen` in text mode strips the `\r` before `fgets` sees it, so CRLF files work. On Linux and macOS the same file leaves a trailing `\r` on the resolved name, so the driver lookup fails and the alias is silently ignored.

Adding `\r` to the delimiter set fixes this in both readers with no behavior change for LF files.

### DIFF
--- a/src/libpinmame/libpinmame.cpp
+++ b/src/libpinmame/libpinmame.cpp
@@ -317,7 +317,7 @@ static const char* CheckGameAlias(const char* const romName)
 
 			if (!strcasecmp(token, romName))
 			{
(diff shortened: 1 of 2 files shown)


toxieainc:
Thanks!

PR #627, opened 2026-08-31
Sleic: Bike Race and Sleic Pin-Ball playable

gerwout:
> **Prerequisite: #625 must be merged before this PR.**
> **Sleic Pin-Ball faults during boot without the `CHANGE_PC` fix in https://github.com/vpinball/pinmame/pull/625, because the 80188 runs its game code at `0xD0000`-`0xFFFFF`. This branch is built on top of that fix so the games are testable today; once #625 is merged the diff here reduces to the driver commit alone.**

### What this does

Brings up three sets on the Sleic boards:

| set | board | before | after |
|---|---|---|---|
| `bikerace` | SLEIC3 | not working | playable |
| `bikerac2` | SLEIC3 | not working | playable |
| `sleicpin` | SLEIC1 | `GAME_NOT_WORKING` | playable |

Io Moon (`iomoon`, SLEIC2) is deliberately **out of scope** — I am still working on this. It needs an undumped PIC16C57 that rasters its DMD, and I would rather submit it when that is solved than ship guesses.

### What works

A coin gives a credit, start begins a game, the ball trough and playfield contacts
register and score, lamps and coils follow the ROM's own tables, the DMD renders, the
service menu is reachable from the test button, settings persist to NVRAM, and both
sound chips play — YM3812 FM music and OKI MSM6376 speech and effects.

### How the boards work

Both use an 80188 for the game, a Z80 for I/O, and a YM3812 plus an OKI MSM6376 for
sound, so the shared parts live in the `SLEIC` machine and each board overrides what
differs.

**Inter-CPU link.** The Z80 and the 80188 talk over the J1 byte port — eight data lines
and handshakes, no shared memory, no HOLD/HLDA. The Z80 writes a switch code and strobes
it; the 80188 takes it from its inbound latch at PCS2 (`0xA0100`). Cabinet buttons arrive
as codes on Z80 port `0x03`, the playfield matrix through the port-`0x82` column strobe.

**Interrupts.** Sleic Pin-Ball's firmware enables only the 80188's internal timer, so
Timer0 has to be delivered on vector `0x08` or the attract mode never runs.

**Reverse path.** PCS1 (`0xA0080`) carries the command byte going back from the 80188 to
the Z80. That is what drives lamp and solenoid updates and lets the flippers navigate the
service menu.

**Sound.** The 80188 owns both chips. The YM3812 is on PCS5, but the two boards address it
differently — Bike Race streams register/value pairs to a single address and toggles A0 in
hardware, Sleic Pin-Ball latches A0 from PCS0 bit 1 — so they are decoded separately. The
OKI takes a phrase number in its control latch and plays it on the channel selected by
bit 7.

**Maps.** Switch and coil assignments follow the service manuals: `sleicpin`'s solenoids
1-12 are bobinas 01-12 and its cabinet contacts are C31-C36. Bike Race's ball-status
handler watches different trough optos per version — `bikerace` reports BOLAS OK from one
of two, `bikerac2` counts them and needs two including its own sensor.

### Testing

Built and run on Linux/SDL. All three sets boot and run at 60 FPS, and all three were
played by hand: trough populated, credits added, game started, playfield contacts and
flippers exercised, sound confirmed on both boards. Bike Race was also checked from a
blank NVRAM through its factory prompt and back.

A few environment variables, inert unless set, help with automated checks:
`SLEIC_DMD_DUMP` writes one PGM per DMD frame, and `SLEIC_TRACE_PW` / `SLEIC_TRACE_SW` /
`SLEIC_TRACE_LAMP` / `SLEIC_TRACE_SND` trace peripheral, switch, lamp and sound writes.
Happy to drop these if you would rather the driver stayed clean of them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ and listen to/record the pinball game sounds with the pure PinMAME package itsel
 - *Sonic* - Odin DeLuxe (1985), Pole Position (1987), Star Wars (1987)
 - *Allied Leisure* - All games from Super Picker (1977) to Star Shooter (1979)
 - *Fascination, Int.* - Roy Clark - The Entertainer (1977), Eros One, and Circa 1933 (both 1979)
-- *Sleic*
+- *Sleic* - Bike Race (1992, standard and 2-ball versions), Sleic Pin-Ball (1993)
 - *Playmatic* - Last Lap (1978), Antar (1979), Evil Fight (1980), Mad Race (1982), Meg-Aaton (1983), KZ-26 (1984) (*)
 - *NSM*
 - *Grand Products* - 300/Bullseye (1986)
--- a/src/cpu/i86/i86.c
+++ b/src/cpu/i86/i86.c
@@ -163,7 +163,7 @@ void i86_reset(void *param)
 	I.pc = 0xffff0 & AMASK;
 	ExpandFlags(I.flags);
 
-	change_pc16(I.pc);
+	change_pc20(I.pc);
 }
 
 void i86_exit(void)
@@ -189,7 +189,7 @@ void i86_set_context(void *src)
 		I.base[DS] = SegBase(DS);
 		I.base[ES] = SegBase(ES);
 		I.base[SS] = SegBase(SS);
-		change_pc16(I.pc);
+		change_pc20(I.pc);
 	}
 }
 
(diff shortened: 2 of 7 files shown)


toxieainc:
@volkenborn Could you please have a look? Thanks!

toxieainc:
Great stuff, thanks!

PR #628, opened 2026-08-31
Keep serving the last display frame after the emulation stops

francisdb:
The DMD and video frame getters returned a null frame once the emulation left the running state, while the display sources stay advertised until retracted on the plugin API thread. Consumers polling in that window crashed (see vpinball/vpinball#3871).

Frames are now copied into libpinmame owned buffers and the last one keeps being served after stop, as the segment display getter already does.

### DIFF
--- a/src/libpinmame/libpinmame.cpp
+++ b/src/libpinmame/libpinmame.cpp
@@ -5,6 +5,7 @@
 #include "../../ext/libsamplerate/samplerate.h"
 
 #include <thread>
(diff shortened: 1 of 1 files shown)


vbousquet:
This was properly fixed by implementing the controller contract (do not advertise something you actually don't provide).

francisdb:
see e48bc7bb8dcd3d6f860d99e68ff6c37185def955

PR #634, opened 2026-09-02
LibPinMAME: write CPU memory through the memory map

jayadelson:
## What

There is no way for a plugin to write the emulated machine's CPU memory today. Adding one is not simply the mirror of the ranged read, because writes carry required platform-specific details that reads do not.

`memory_get_write_ptr()` returns NULL wherever the write table holds a handler rather than a direct pointer. For example, on WPC that is all of `0x0000-0x2fff`, so a pointer-based write silently no-ops on CMOS. Where it does return a pointer, writing through it bypasses whatever the handler was doing:

| Platform | Handler | What a pointer write gets wrong |
|---|---|---|
| Williams WPC | `wpc_ram_w` | Bypasses the memory protection register |
| Bally (early) | `by35_CMOS_w` | Stores 8 bits where the CMOS holds 4 |
| Gottlieb System 80 | `ram_256w` | Updates 1 of 32 mirrors, leaving 31 stale |

Two of those three fail silently, as the write reports success and stores a value the machine could never contain.

## Why a sequence rather than a single range

Lessons learned doing this on real hardware... Writing protected CMOS on WPC means storing `0xB4` to `0x3FFD` first, and the game re-locks within a frame. Two separate messages are a frame apart, so the payload arrives after the re-lock and is, therefore, dropped.

Same binary, same address, six seconds into attract mode on `ij_l7`, only the grouping differs:

```
unlock + payload in one message    read observes the write
unlock and payload as two          write is refused
```

So the message carries a list of ops applied in a _single pass_ rather than one range. If you would rather have a different mechanism, I am happy to rework it... The only requirement is that the ops are indivisible with respect to the emulated CPU.

## Implementation

Shamelessly queued and applied from `osd_update_video_and_audio()`, which runs once per frame on the emulation thread, between CPU time slices, where the active-CPU context is valid and handler dispatch is legal. It did precisely what I needed. `cpunum_write_byte()` is not safe to call from the asynchronous message thread though.`libpinmame.cpp` already notes that switching CPU context there is a pain not worth experiencing.

Everything drained in one pass is atomic with respect to the machine, which is what replaces halting the CPU. A caller blocks until the next drain, measured at 9.5 ms median and 16.5 ms maximum, bounded at 250 ms so a paused emulator cannot hang a plugin.

`PinmameWriteMainCPUMemory` follows the merged ranged read, with the buffer const since it is only read from.

## Testing

Built on macOS/arm64. Read-modify-write-verify of 8 bytes of main RAM, six seconds into attract mode, restoring the original contents afterwards:

`ij_l7` (WPC, protection unlock), `flashgdn` (Bally 4-bit, nibble mask observed), `blckhole` (Gottlieb System 80, mirror confirmed at `0x1900`), `whirl_l3` (System 11), `cueball` (Gottlieb System 3, multi-CPU) — all pass.

Also exercised end to end in VPX 10.8.1 Beta with a small test plugin loaded alongside PinMAME, ScoreView, B2SLegacy, AltSound, PUP, Serum, FlexDMD and VNI: an unprotected write is refused, the batched write lands, and the teardown is clean.

The test harness is out of this since nothing in CI runs it, but I can add it under `src/libpinmame/` if that is useful! I see there is precedent in `src/remote_debug/test_suite.sh`. Just let me know.

## Disclosure

### DIFF
--- a/src/libpinmame/PinMAMEPlugin.h
+++ b/src/libpinmame/PinMAMEPlugin.h
@@ -45,6 +45,26 @@ typedef struct PinMAMEReadMemoryMsg
    uint32_t read;       // Response: bytes actually read, 0 if unavailable
 } PinMAMEReadMemoryMsg;
 
+#define PMPI_WRITE_MEMORY                    "WriteMemory:1"
+
+typedef struct PinMAMEWriteMemoryOp
+{
+   uint32_t address;       // Start address in the main CPU address space
+   uint32_t size;          // Number of bytes to write
+   const uint8_t* data;    // Caller-owned buffer, at least 'size' bytes
+} PinMAMEWriteMemoryOp;
+
+// Every op in one message is applied in a single pass on the emulation thread, so the
+// machine cannot run between them. WPC needs that: writing protected CMOS means storing
+// 0xB4 to 0x3FFD first, and the game re-locks within a frame.
+typedef struct PinMAMEWriteMemoryMsg
+{
+   int version;                      // Always 1 (to allow upgrading this message in later revisions)
+   uint32_t count;                   // Request: number of ops
+   const PinMAMEWriteMemoryOp* ops;  // Request: applied in order, atomically
+   uint32_t applied;                 // Response: count on success, 0 if unavailable
+} PinMAMEWriteMemoryMsg;
+
 
 // State groups
 // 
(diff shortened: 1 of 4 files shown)


jayadelson:
Rebased onto master so the message is now `WriteMemory:1` to match the versioned naming from 3215a8ea.

jayadelson:
@toxieainc any changes you need on our end?

PR #348, opened 2024-10-21
feat: sdl3

francisdb:
I ported the SDL code to SDL3. (continuation of #203)

Used the approach of building SDL3 in the project like is done on standalone / libdmdutil / libaltsound. SDL3 is currently not available in homebrew nor ubuntu apt repositories.

X11 is ancient tech that is going away.
`xpinmame` on linux now builds against SDL3 for graphics and sound.

Some light crackling on the sound, probably something to do with pinmame not coming up with audio data in time? Had the same issue with coreaudio.

```
SDL_AUDIODRIVER=pipewire SDL_VIDEODRIVER=wayland ./build/xpinmame -rompath ~/.pinmame/roms -nvram_directory ~/.pinmame/nvram  t2_l8
```

image

TODO

- [x] Multi platform builds (mac/linux)
- [x] Support for HiDPI (scale up window contents)? For now we use actual pixels on macOS. (tiny window, requires `-scale 4` for example)
- [x] Sound not working on mac? (somehow main speakers don't work with coreaudio, switched to sdl3 for sound)
- [ ] Probably needs the SDL3 lib in the archive
- [x] ~~does the -md build still work and show the debug window, was this ever the case?~~ (not in use for xpinmame)

PR #640, opened 2026-09-02
sdl3pinmame: command line PinMAME using SDL3 for video, input and sound

francisdb:
Supersedes #348, restarted from scratch on current master as a new `sdl3pinmame` target instead of changing xpinmame.

- New SDL3 display/input/joystick driver and SDL3 sound plugin (`src/unix/video-drivers/sdl3.c`, `src/unix/sysdep/dsp-drivers/sdl3.c`), separate files so the SDL2 based lisy build is untouched.
- One platform-agnostic cmake file (`cmake/sdl3pinmame/CMakeLists.txt`, `-DPLATFORM=linux|macos -DARCH=x64|arm64`) that fetches and statically links SDL 3.4.14 by default (`-DSDL3PINMAME_SYSTEM_SDL=ON` to use an installed one), plus a workflow for linux-x64, macos-x64 and macos-arm64.
- Optional `-DSDL3PINMAME_MAME_DEBUG=ON` build with the MAME debugger in its own window next to the game window.
- The cmake file also avoids two pitfalls inherited from the xpinmame cmake files: the unquoted `INLINE=static __inline__` define (turns `__inline__` into `1`) and the missing `HAVE_VSNPRINTF` (the vsnprintf replacement in `src/unix/snprintf.c` crashes on `vsnprintf(NULL, 0, ...)`).

Tested on Linux (Wayland and XWayland, pipewire), macOS builds are CI only so far.

### DIFF
--- a/.github/workflows/sdl3pinmame.yml
+++ b/.github/workflows/sdl3pinmame.yml
@@ -1,0 +1,85 @@
+name: sdl3pinmame
+on:
+  push:
+  pull_request:
+
+env:
+  VERSION_START_SHA: 79345956d1807e3188f0e895379466c7c4caae72
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  version:
+    name: Version
+    runs-on: ubuntu-24.04
+    outputs:
+      version: ${{ steps.version.outputs.version }}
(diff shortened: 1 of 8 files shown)


toxieainc:
Nice, thanks!

PR #643, opened 2026-09-02
LibPinMAME: read CPU memory on the emulation thread

### LINKED ISSUE #642: LibPinMAME: PinmameReadMainCPUMemory intermittently returns zeros for a whole chunk while reporting a full read
## What happens

I'm hoping you can help with this. `PinmameReadMainCPUMemory()` (which was merged in #623) called from a thread other than the emulation thread intermittently fills a 256-byte chunk with zeros and returns `size`, so the caller cannot tell. It happens on fixed ROM, so banking didn't explain it. `PMPI_READ_MEMORY` (#624) inherits it because it calls the same function on the message sender's thread.

## Reproduction, on unmodified master `6589fc57`

I used ROM `ij_l7`, range `0x8000-0xFFFF` read in `0x100` chunks from the main thread, 3 s after the game reaches running, with **50 ms between chunks**. Five walks per run; a chunk is "zeroed" if all 256 bytes are zero and "differing" if it differs from the same chunk in walk 0.

| build | spacing | emulation | zeroed chunks | chunks differing from walk 0 |
|---|---|---|---|---|
| master `6589fc57` | 50 ms | running | 4 of 5 walks hit (`a800`, `b100`, `d700`, `f900`) | 11 |
| master `6589fc57` | 50 ms | paused (`PinmamePause(1)` around each walk) | 0 | 0 |
| master `6589fc57` | none (back to back) | running or paused | 0 in 16 walks | 0 |

I had a friend use his harness on the same master build, same shape but one walk per process in five separate processes, so any in-process state accumulation is ruled out: 5 of 5 running walks wrong, 3 to 8 zeroed chunks each; 5 of 5 paused walks correct. The affected set differs substantially between walks and no walk's set is contained in another's, but individual addresses do recur: across 26 observations there were 21 distinct addresses, with `9b00` hit in three of the five walks and `9000`, `a800` and `ac00` in two each (`8300 8400 8b00 9000 9b00 a000 a800 da00` / `9600 9b00 ac00 ad00` / `a100 a200 ac00 b100` / `8600 9c00 a800` / `8500 8f00 9000 9400 9a00 9b00 b400`). Some addresses come up more than others, so it looks less like a uniform race and more like a short window where part of the map resolves wrong. I hope this makes sense. The two harnesses also overlap: `a800` and `b100` were hit by both, `a800` in two of the second harness's walks.

And, as it turns out... the spacing matters a lot. 16 back-to-back walks on master never hit it (32 including a build carrying #634), while walks spread over ~6 s hit it nearly every time, so the bad state is something the emulation enters periodically and briefly rather than a uniform race. Importantly, I think pausing removes it.

Tool: `read_walk.cpp` ([read_walk.cpp](https://github.com/user-attachments/files/31757036/read_walk.cpp)) links only libpinmame. `read_walk ij_l7 0x8000 0xFFFF 5 3 --delay=50` reproduces; add `--pause` for the control.

## Likely cause

Here's my supposition here, but I'm new at this. I've asked Claude and a few tools to help me identify why this might be happening. `memory_get_read_ptr()` resolves through `cpu_bankbase[]`, which `memory_set_context()` rewrites to the active CPU's bases on every CPU switch (`memory.c`).

A pointer resolved on another thread, as a result, points into whichever CPU happens to be active at that instant; when it is the sound CPU, the main CPU's ROM addresses resolve to something else, and on WPC that reads as zeros. The same pointer path also has nothing to return for handler-backed RAM (Whitestar and Sega main RAM behind `ram_r`, see #633), which is the short-read half of the same problem.

## Fix

IMHO, we should apply reads on the emulation thread between time slices, the way #634 applies writes, using `cpunum_read_byte()` so the read dispatches through the memory map like a CPU read. That removes both the cross-CPU resolution (the context is valid between slices) and the handler-backed short read. The caller on another thread waits for the next frame, bounded, the same as a write.

I've implemented this and measured it on branch `read-memory-emulation-thread` (PR #643 , stacked on #634 because it reuses its drain). On that branch the reproduction above gives 0 zeroed and 0 differing chunks in 5 walks, and the second harness gives the reference signature 5 of 5 unpaused where master gave 5 different wrong ones; the Whitestar read that #633 addresses (`austin 0x1657`, main RAM behind `ram_r`) returns 8/8 without #633's fallback, so #633 can be closed in favour of it; #634's write harness is unchanged. Cost is one frame of latency per read call from another thread, about 14 ms at 60 FPS; a batched read message is the natural follow-up for callers that need many ranges.

## Disclosure

jayadelson:
Fixes #642. Rebased onto `2006ce5b` (which merged #633); replaces the pointer-path read, including the `memory_find_base` fallback #633 added, with the emulation-thread read.

## What

`PinmameReadMainCPUMemory()` and `PinmameReadMainCPUByte()` resolved a pointer with `memory_get_read_ptr()` on the caller's thread. That has two failure modes, both silent to the caller:

- The bank tables that pointer goes through are switched by `memory_set_context()` every time the emulation moves between CPUs. Resolved on another thread, the pointer points into whichever CPU is active at that instant. On `ij_l7`, reading fixed ROM `0x8000-0xFFFF` in 256-byte chunks with 50 ms between chunks returns a whole chunk of zeros in most walks, with the full count. Details, numbers and a standalone tool are in #642.
- Handler-backed RAM has no pointer to return. Whitestar and Sega route all of main RAM through `ram_r`, so every read there came back short; #633 papered over that with a `memory_find_base` fallback, which this change removes because the emulation-thread read dispatches through `ram_r` itself.

This change queues reads and applies them in the per-frame drain #634 added, on the emulation thread between time slices, with `cpunum_read_byte()` under the main CPU's context, so a read goes through the memory map exactly as the CPU's own would. Writes are applied before reads in the same drain so a read queued after a write observes it. A caller on the emulation thread reads directly. A caller elsewhere waits for the next frame, bounded at 250 ms like a write. A read is short only when the range runs past the end of the address space. `PMPI_READ_MEMORY` gets the same behaviour through the same function; its header comment now says so.

Everything is inside `src/libpinmame/`; no core file changes. The drain is renamed from `libpinmame_drain_pending_writes` to `libpinmame_drain_pending_memory_ops`.

Cost: a read from another thread now costs one frame of latency instead of returning immediately. Measured here: 384 chunk reads in about 5.4 s, so roughly 14 ms each at 60 FPS. A caller that needs many ranges per frame should batch; a message carrying a list of ranges, like `PMPI_WRITE_MEMORY` already does, is the natural follow-up and I would rather propose it separately than widen this one.

Reading a register through its handler has whatever side effect the hardware has. The callers this serves read RAM, NVRAM and ROM; the previous pointer path never reached registers at all, so this is new exposure only for a caller that asks for one.

## Verified, on this branch

| check | result |
|---|---|
| `read_walk ij_l7 0x8000 0xFFFF 5 3 --delay=50`, the reproduction shape | 5 walks, 0 zeroed chunks, 0 differing chunks (was 4 of 5 walks zeroed on master) |
| `read_test austin 0x1657 8`, Whitestar main RAM through `ram_r` (#633's case) | `ranged=8/8 byte=1`, without #633's fallback |
| `write_memory_test --rom ij_l7 --addr 0x1c93 --len 8 --settle 6 --unlock 0x3FFD 0xB4` | 11 passed, 0 failed, unchanged from #634 |
| same with `--split` (unlock and payload as two messages) | fails the read-back as before, which is the behaviour #634 documents |
| independent second harness (firmware-signature walk, same shape, unpaused) | 5 of 5 walks give the reference signature `e029a575`, 0 zeroed chunks; the same configuration on master gave five different wrong signatures |

Tools are the ones offered on #634 plus `read_walk.cpp`; happy to add them under `src/libpinmame/` if wanted.

## Disclosure

### DIFF
--- a/src/libpinmame/PinMAMEPlugin.h
+++ b/src/libpinmame/PinMAMEPlugin.h
@@ -45,6 +45,26 @@ typedef struct PinMAMEReadMemoryMsg
    uint32_t read;       // Response: bytes actually read, 0 if unavailable
 } PinMAMEReadMemoryMsg;
 
+#define PMPI_WRITE_MEMORY                    "WriteMemory:1"
+
+typedef struct PinMAMEWriteMemoryOp
+{
+   uint32_t address;       // Start address in the main CPU address space
+   uint32_t size;          // Number of bytes to write
+   const uint8_t* data;    // Caller-owned buffer, at least 'size' bytes
+} PinMAMEWriteMemoryOp;
+
+// Every op in one message is applied in a single pass on the emulation thread, so the
+// machine cannot run between them. WPC needs that: writing protected CMOS means storing
+// 0xB4 to 0x3FFD first, and the game re-locks within a frame.
+typedef struct PinMAMEWriteMemoryMsg
+{
+   int version;                      // Always 1 (to allow upgrading this message in later revisions)
+   uint32_t count;                   // Request: number of ops
+   const PinMAMEWriteMemoryOp* ops;  // Request: applied in order, atomically
+   uint32_t applied;                 // Response: count on success, 0 if unavailable
+} PinMAMEWriteMemoryMsg;
+
 
 // State groups
 // 
--- a/src/libpinmame/PinMAMEPlugin.h
+++ b/src/libpinmame/PinMAMEPlugin.h
@@ -36,6 +36,11 @@ typedef struct PinMAMEMachineStateMsg
 
 #define PMPI_READ_MEMORY                     "ReadMemory:1"
 
+// The read is performed on the emulation thread between time slices, dispatched through the
+// memory map like a CPU read, so handler-backed RAM reads correctly and the result cannot be
+// torn by the running emulation or resolved against another CPU's bank tables. The caller
+// waits for the next frame (bounded, typically under 20 ms). 'read' is short only when the
+// range runs past the end of the address space.
 typedef struct PinMAMEReadMemoryMsg
 {
    int version;         // Always 1 (to allow upgrading this message in later revisions)
--- a/src/libpinmame/libpinmame.cpp
+++ b/src/libpinmame/libpinmame.cpp
@@ -1753,64 +1753,6 @@ PINMAMEAPI int PinmameGetChangedNVRAM(PinmameNVRAMState* const p_nvramStates)
 }
 
 /******************************************************
- * PinmameReadMainCPUByte
- ******************************************************/
-
-PINMAMEAPI int PinmameReadMainCPUByte(const uint32_t address, uint8_t* const p_value)
-{
-	return PinmameReadMainCPUMemory(address, p_value, 1);
-}
-
-/* A NULL from memory_get_read_ptr() means the read table holds a handler rather than
-   a direct pointer, not that the address is unreadable at all, se.c routes all of
-   Whitestar/Sega main RAM through ram_r. memory_find_base() has no bounds check. */
-
-static int ReadThroughRamBase(const uint32_t address, uint8_t* const p_buffer, const int size)
-{
-	const size_t regionLength = memory_region_length(REGION_CPU1);
-	if (address >= regionLength)
-	{
-		return 0;
-	}
-
-	const uint8_t* const p_base = static_cast<const uint8_t*>(memory_find_base(0, address));
-	if (p_base == nullptr)
-	{
-		return 0;
-	}
-
-	const int available = (int)std::min((size_t)size, regionLength - address);
-	memcpy(p_buffer, p_base, available);
-
-	return available;
-}
-
-/******************************************************
- * PinmameReadMainCPUMemory
- ******************************************************/
-
-PINMAMEAPI int PinmameReadMainCPUMemory(const uint32_t address, uint8_t* const p_buffer, const int size)
-{
-	if (!_isRunning || p_buffer == nullptr || size <= 0)
-	{
-		return 0;
-	}
-
-	for (int i = 0; i < size; i++)
-	{
-		uint8_t* p_memory = static_cast<uint8_t*>(memory_get_read_ptr(0, address + i));
-		if (p_memory == nullptr)
-		{
-			return i > 0 ? i : ReadThroughRamBase(address, p_buffer, size);
-		}
-
-		p_buffer[i] = *p_memory;
-	}
-
-	return size;
-}
-
-/******************************************************
  * PinmameWriteMainCPUMemory
  ******************************************************/
 
@@ -1823,9 +1765,21 @@ struct PendingWrite
 	bool done;
 };
 
+struct PendingRead
+{
+	uint32_t address;
+	uint8_t* buffer;
+	int size;
+	int read;
+	bool done;
+};
+
+/* One mutex and one condition variable serve both queues: they are drained together, writes
+   first, so a read queued after a write observes it. */
 std::mutex g_writeMutex;
 std::condition_variable g_writeCv;
 std::vector<PendingWrite*> g_writeQueue;
+std::vector<PendingRead*> g_readQueue;
 
 } /* anonymous namespace */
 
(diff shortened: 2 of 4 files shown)


toxieainc:
Somehow i wonder why this whole architecture should work like this in the first place. Wouldn't it make more sense to be able to just read/write each emulated CPUs memory region independently without jumping through hoops?

jayadelson:
> Somehow i wonder why this whole architecture should work like this in the first place. Wouldn't it make more sense to be able to just read/write each emulated CPUs memory region independently without jumping through hoops?

Yeah... @toxieainc you're right.

Honestly, the API already looks like it reads each CPU's memory independently, and doesn't.

`memory_get_read_ptr(cpunum, offset)` resolves the entry from the requested CPU's own read table, then dereferences `cpu_bankbase[entry]`, which is global, and `memory_set_context()` rewrites `cpu_bankbase[STATIC_RAM]` to the *active* CPU's `rambase` on every context switch.

On an 8-bit CPU every RAM and ROM pointer read goes through that one entry (the guard above the return only lets `STATIC_RAM` through when `minbits == 0`). So... a main-CPU address resolved while the sound CPU is active lands in the sound CPU's region, and on WPC that reads as zeros. Which is annoying. That's the whole mechanism behind #642, which is the right table, wrong base.

So there are two ways to make it per-CPU, and I've taken a look at both:

1. **We make the base per-CPU in `memory_get_read_ptr`**: for the `STATIC_RAM` entry, return `cpudata[cpunum].rambase + offset` instead of `cpu_bankbase[STATIC_RAM]`. We keep it one line in memory.c, and reads stay on the caller's thread. On the #642 reproduction (ij_l7, 0x8000-0xFFFF, 50 ms between chunks, unhalted) that gives 0 zeroed chunks in 15 walks (master gives 4 or 5 of 5 walks with zeroed chunks), so the wrong-CPU resolution is the whole cause of the zeros. It fixes the wrong-CPU resolution and nothing else, though. A multi-byte read still runs concurrently with the emulation and can tear, and handler-backed regions still need #633's fallback, which reads the raw region and so bypasses whatever the handler does (Whitestar's `ram_r` exists because of TRACERAM banking).
2. **We can do the read on the emulation thread between slices, through the CPU's own read path** (this PR). Reuses the queue #634 added for writes. Gives 0 zeroed chunks in 5 walks and reads Whitestar main RAM through `ram_r` without the fallback. The cost is a frame of latency per call for a caller on another thread, about 14 ms, which is fine.

I went with 2 because it's local to libpinmame and changes no core semantics, and because it also answers the tearing and handler questions... But I just got here... if you'd rather fix `memory_get_read_ptr` **properly***, LFG!!!! That's the better long-term solution and I'd take it over this PR. The one-line version above is the "minimal" form of it, and per-CPU bank bases across the board would be the full form. Either way a caller reading a running emulator from another thread still needs _something_ to synchronise against for multi-byte reads, which is the part option 1 doesn't give.

Happy to change this PR to whichever you prefer, just let me know!

jayadelson:
@toxieainc @vbousquet Short version: you're right, and the one-line fix in memory.c works too.

`memory_get_read_ptr` uses the requested CPU's table but the global `cpu_bankbase[STATIC_RAM]`, which the context switch repoints to the active CPU. Returning `cpudata[cpunum].rambase + offset` for that entry kills the zeros (0 in 15 walks vs 4-5 of 5 on master).

That's a smaller change than this PR. So my question really is: want me to swap this PR for that one-liner, or keep the emulation-thread read? Either is fine by me.

jayadelson:
Rebased onto master so the message is now `WriteMemory:1` to match the versioned naming from 3215a8ea.

jayadelson:
Closing in favor of the direction #685 as @toxieainc suggested, and it leaves one read fix to review instead of two.
@kylehudson00, that should let #687 drop the emulation-thread dispatch it adapted from this PR. On your question: a raw pointer only covers regions backed by memory. Handler-backed regions still need the handler, which is why #633  falls back to it for `PinmameReadMainCPUMemory,` so I'd keep `cpunum_read_byte()` semantics for those.

Issue #608, opened 2026-08-15
Goldeneye solenoid 33 is activating the right flipper solenoid. It should not.


apophis79:
**Describe the bug**
When playing VPW's Goldeneye version 1.2, the SolCallback(33)="SolRadarMagnet" also activates the "SolRFlipper" when it should not. I am not sure when this problem started.

**To Reproduce**

1. Download Goldeneye version 1.2 from https://vpuniverse.com/files/file/10040-goldeneye-sega-1996-vpw-mod/
2. Instrument the script with a debug.print inside the SolRadarMagnet sub to check whether it gets called
3. Instrument the script with a debug.print inside the SolRFlipper sub to check whether it gets called
4. Run the game with the debug window open and use ball control to start satellite multiball: 
Shoot the standup target in the back center of the table to light a lock. 
Locks are collected at the center ramp and are entirely virtual. Two locks are needed to light the left inlane.
Drop the ball down the left inlane to activate the radar magnet.
5. Notice that the right flipper activates when it should not. Look for the debug.print in the debugger window and you call see the solenoids are both getting called.

**Expected behavior**
The SolRadarMagnet callback should not activate SolRFlipper

**Versions & Settings**
VPinMame: [VPinMAME-3.7.0-403-2254e45-win-x64](https://github.com/vpinball/pinmame/actions/runs/31884291197/artifacts/9246903788)
VPX: 10.8.0-2051 windows directX 64-bit
Windows version 11

gitfool:
I reproduced this on a real GoldenEye 1.2 table and instrumented the build to trace it at the C layer. The right flipper does physically engage with the button untouched, so this is a real bug. But the data says the radar magnet is coincidental, not the cause. Here is what is actually happening.

Tested on vpinball [d363edb18](https://github.com/vpinball/vpinball/commit/d363edb180b8219b4ddd5ec94f87932141efb3b6) (macOS BGFX) with its pinned PINMAME_SHA [959c58bb](https://github.com/vpinball/pinmame/commit/959c58bb41012b24e870dead04630f25039e9957), the current master fix from #645. The only change to that dylib was two revert-only diagnostic probes I added to capture the traces below, so the behavior is not an artifact of the instrumentation.

### The magnet and the right flipper are independent in the solenoid state

[...]

PR #645, opened 2026-09-03
LibPinMAME: report flipper solenoids full scale in modsol/physout mode

gitfool:
Found while looking into #608 (GoldenEye's radar magnet also firing the right flipper). I never got to that bug, because on the current build GoldenEye's flippers do not respond at all, so there was nothing to reproduce. This PR fixes the dead flippers.

## What is wrong

`GetSolenoid2VPMState` reports an "on" binary solenoid as `1`. When a table enables physical output (`UseVPMModSol = 2`), `core.vbs` normalizes each changed-solenoid value by the group's 0..255 range (`pwmScale = 1/255`) and thresholds at `>= 0.5`. A reported `1` becomes about `0.004` and reads as off.

Most solenoids dodge this. They either report through the physical-output path (`GetPhysOutVPMState`, already 0..255) or are consumed where `pwmScale` is `1`. The flipper coils (solenoids 33-36 and 45-48) are the exception. On hardware without modulated flippers (`hasModulatedFlippers == false`: Whitestar, System 11, Data East, pre-Fliptronic WPC, and older generations) they are read from the raw `solenoids2` bitfield through `GetSolenoid2VPMState`. If a table drives its flipper enable through one of those coils and also turns on physical output, the enable collapses to off and the flippers never engage.

## When it regressed

This came in with 04b2da8e ("LibPinMAME: update to new Controller plugin API", 2026-08-24), which moved solenoid state onto the new controller-plugin state groups. The old path (`vp_getChangedSolenoids`) ran every solenoid through `saturatedByte` (0..255) in physical mode, with no per-coil exception, so the flipper enable came through full-scale. The new `GetSolenoid2VPMState` reports binary `1` for the non-modulated flipper coils instead.

## Reproduction

GoldenEye (Sega Whitestar). `sega2.vbs` sets `GameOnSolenoid = 48`, the lower-left flipper Hold|Power coil, and the table sets `UseVPMModSol = 2`. The enable solenoid arrived as `1`, scaled to `0.004`, so the flipper circuit never switched on and the flippers were dead with the stock, unmodified script.

## Fix

Report `0xFF` for an "on" binary solenoid so the value survives the consumer's 0..255 normalization. This is the convention `GetSwitchState` and `GetDIPSwitchState` already follow.

The `hasModulatedFlippers` gate stays as is. I tried removing it so the flipper coils would take the physical-output path like everything else, but that reads from a PWM model that is not driven for these coils on non-modulated hardware, so they report nothing and the flippers go dead a different way. The gate is load-bearing. `GetSolenoid1VPMState` carries the same `1`-for-on pattern but is only ever reached through the scale-consistent `addPhysSol` path, so it has no reachable defect and I left it alone.

## Testing

Flip-tested stock tables on macOS/BGFX standalone with the PinMAME plugin.

| Table | Generation, GameOn solenoid | Before | After |
| --- | --- | --- | --- |
| GoldenEye (Sega 1996) | Whitestar, sol 48, `UseVPMModSol=2` | dead | works |
| Jurassic Park (Data East 1993) | sol 23, `UseVPMModSol=2` | works | works |
| Twister (Sega 1996) | Whitestar, sol 15 | works | works |
| Medieval Madness (Williams 1997) | Fliptronic WPC | works | works |

The three regression cases cover the other affected generations, the two other `GameOnSolenoid` values in common use (15 and 23), and a modulated-flipper generation on the untouched physical path.

### DIFF
--- a/src/libpinmame/libpinmame.cpp
+++ b/src/libpinmame/libpinmame.cpp
@@ -1899,6 +1899,14 @@ static void GetSolenoid2VPMState(void* callContext, void* pResult)
       core_update_pwm_outputs(CORE_MODOUT_SOL0 + core_BitColToNum(srcId) + 32, 1);
    *static_cast<uint8_t*>(pResult) = (coreGlobals.solenoids2 & srcId) != 0 ? 1 : 0;
 }
+static void GetFlipperSolenoid2VPMState(void* callContext, void* pResult)
+{
+   assert(_isRunning == 1);
+   const int srcId = static_cast<StateMapping*>(callContext)->srcId;
+   if (options.usemodsol & CORE_MODOUT_FORCE_ON)
+      core_update_pwm_outputs(CORE_MODOUT_SOL0 + core_BitColToNum(srcId) + 32, 1);
+   *static_cast<uint8_t*>(pResult) = (coreGlobals.solenoids2 & srcId) != 0 ? 0xFF : 0;
+}
 static void GetCustomSolenoidState(void* callContext, void* pResult)
 {
    assert(_isRunning == 1);
@@ -2166,8 +2174,9 @@ static void SetupMsgApiGameStates()
                   case 35:mask = isFlipperCoil ? 0x40 : 0x40; break; // Power bit
                   case 36:mask = isFlipperCoil ? CORE_ULFLIPSOLBITS : 0x80; break; // Power|Hold bits
                   }
+                  const bool physOut = (options.usemodsol & (CORE_MODOUT_ENABLE_PHYSOUT_SOLENOIDS | CORE_MODOUT_ENABLE_MODSOL | CORE_MODOUT_FORCE_ON)) != 0;
                   addDevice(PMPI_GROUP_SOLENOID, label, nullptr, i, CTLPI_STATE_FORMAT_FLOAT, CTLPI_STATE_TYPE_CUSTOM, GetSolenoid2State, nullptr, mask);
-                  addDevice(PMPI_GROUP_VPM_SOLENOID, fmtString("%s", label), nullptr, i, CTLPI_STATE_FORMAT_UINT8, CTLPI_STATE_TYPE_CUSTOM, GetSolenoid2VPMState, nullptr, mask);
+                  addDevice(PMPI_GROUP_VPM_SOLENOID, fmtString("%s", label), nullptr, i, CTLPI_STATE_FORMAT_UINT8, CTLPI_STATE_TYPE_CUSTOM, (isFlipperCoil && physOut) ? GetFlipperSolenoid2VPMState : GetSolenoid2VPMState, nullptr, mask);
                }
             }
          }
@@ -2220,8 +2229,9 @@ static void SetupMsgApiGameStates()
             case 47:mask = 0x04; break; // Power bit
             case 48:mask = CORE_LLFLIPSOLBITS; break; // Power|Hold bits
             }
+            const bool physOut = (options.usemodsol & (CORE_MODOUT_ENABLE_PHYSOUT_SOLENOIDS | CORE_MODOUT_ENABLE_MODSOL | CORE_MODOUT_FORCE_ON)) != 0;
             addDevice(PMPI_GROUP_SOLENOID, label, nullptr, i, CTLPI_STATE_FORMAT_FLOAT, CTLPI_STATE_TYPE_CUSTOM, GetSolenoid2State, nullptr, mask);
-            addDevice(PMPI_GROUP_VPM_SOLENOID, fmtString("%s", label), nullptr, i, CTLPI_STATE_FORMAT_UINT8, CTLPI_STATE_TYPE_CUSTOM, GetSolenoid2VPMState, nullptr, mask);
+            addDevice(PMPI_GROUP_VPM_SOLENOID, fmtString("%s", label), nullptr, i, CTLPI_STATE_FORMAT_UINT8, CTLPI_STATE_TYPE_CUSTOM, physOut ? GetFlipperSolenoid2VPMState : GetSolenoid2VPMState, nullptr, mask);
          }
       }
       // 49, simulated fake plunger, not broadcasted


vbousquet:
The bug analyze is right but the proposed fix is actually wrong.

The problem is the conversion of this (already not that clean) code from core_getAllPhysicSols:
```
  /*-- 37..44, extra solenoids --*/
  if (core_gameData->gen & (GEN_WPC95 | GEN_WPC95DCS)) // 37-44 WPC95 extra (duplicated 37..40 / 41..44)
    for (int i = 36; i < 40; i++) {
      state[i    ] = coreGlobals.physicOutputState[CORE_MODOUT_SOL0 + i].value;
      state[i + 4] = coreGlobals.physicOutputState[CORE_MODOUT_SOL0 + i].value;
    }
  else if (core_gameData->gen & (GEN_ALLS11 | GEN_SAM | GEN_SPA)) // 37-44 S11, SAM extra
    for (int i = 40; i < 48; i++)
      state[i - 4] = coreGlobals.physicOutputState[CORE_MODOUT_SOL0 + i].value;
  /*-- 45..48 lower flipper solenoids --*/
  if (coreGlobals.hasModulatedFlippers && (options.usemodsol & (CORE_MODOUT_ENABLE_PHYSOUT_SOLENOIDS | CORE_MODOUT_ENABLE_MODSOL | CORE_MODOUT_FORCE_ON)))
    for (int i = 44; i < 48; i++)
      state[i] = coreGlobals.physicOutputState[CORE_MODOUT_SOL0 + i].value;
  else {
    UINT8 lFlip = (coreGlobals.solenoids2 & (CORE_LRFLIPSOLBITS|CORE_LLFLIPSOLBITS));
    lFlip |= (lFlip & 0x05)<<1; // hold coil is set if either coil is set
    state[44] = lFlip & 0x01 ? 1.0f : 0.0f;
    state[45] = lFlip & 0x02 ? 1.0f : 0.0f;
    state[46] = lFlip & 0x04 ? 1.0f : 0.0f;
    state[47] = lFlip & 0x08 ? 1.0f : 0.0f;
  }
```

The `if (core_gameData->gen & (GEN_ALLS11 | GEN_SAM | GEN_SPA)) // 37-44 S11, SAM extra` has not been ported and need to be like it is for upper flipper, using addPhysSol in libPinMAME. Adding it would solve the bug properly

gitfool:
Thanks. I tried the `addPhysSol` route for the lower flippers to check I understood you:

```cpp
// 45..48, lower flipper solenoids
for (uint16_t i = 45; i < 49; i++)
{
   /* label unchanged */
   int mask;
   switch (i)
   {
   case 45: mask = 0x01; break;
   case 46: mask = CORE_LRFLIPSOLBITS; break;
   case 47: mask = 0x04; break;
   case 48: mask = CORE_LLFLIPSOLBITS; break;
   }
   addPhysSol(label, nullptr, nullptr, i, GetSolenoid2State, GetSolenoid2VPMState, mask, i - 1);
}
```

GoldenEye flippers are still dead with this (Whitestar, `GameOnSolenoid = 48`, `UseVPMModSol = 2`).

`isPhysSol` is true, so `addPhysSol` reads `physicOutputState[47]`, which isn't driven for flipper coils on hardware without modulated flippers, so it stays 0. `core_getAllPhysicSols` handles that same case by reading `solenoids2` (with hold-coil synthesis) and reporting 1.0f, not `physicOutputState`. The else branch of `addPhysSol` does read `solenoids2`, but only through `GetSolenoid2VPMState`, which is the binary `1` we started with.

So for these coils the value has to come from `solenoids2` at full scale. Sourcing them there and reporting `0xFF` for on (matching `saturatedByte(1.0f)`) fixes GoldenEye and leaves `GetSolenoid2VPMState` binary for the ext-board and GPIO callers. Tested clean on Jurassic Park (sol 23), Twister (sol 15), and Medieval Madness.

Did I read your suggestion right, or did you have different `addPhysSol` wiring in mind?

vbousquet:
Sorry, my quick review was not clear and in fact somewhat wrong (the codeblock to port is right, the hint regarding the addPhysSol is wrong). What I mean is that we are supposed to reproduce what was done by vp_getChangedSolenoids to be fully backward compatible.

So it is core_getAllPhysicSols when using modsol or physout (0..255 output) and core_getAllSol otherwise (0..1 output). The initial fix did not guaranteed this as it would always return 255 (so likely breaking tables expecting 1 instead of 255, like the bug we had at some point with lamps on a few tables).

The above codeblock from core_getAllPhysicSols  shows that the result should 1.f / 255 when in modsol/physout only, so this must be implemented like this. I guess the easiest correct way if to have 2 variants of GetSolenoid2VPMState corresponding to these 2 situations and using the right one when advertising the state

gitfool:
Thanks, that's clear now. Done that way.

Added `GetFlipperSolenoid2VPMState` (returns `0xFF` for on) alongside `GetSolenoid2VPMState` (unchanged, binary), and the flipper blocks pick between them by mode:

```cpp
const bool physOut = (options.usemodsol & (CORE_MODOUT_ENABLE_PHYSOUT_SOLENOIDS | CORE_MODOUT_ENABLE_MODSOL | CORE_MODOUT_FORCE_ON)) != 0;
...
physOut ? GetFlipperSolenoid2VPMState : GetSolenoid2VPMState
```

So modsol/physout reports 0..255 (255 for on, matching `core_getAllPhysicSols` -> `saturatedByte(1.0f)`), and everything else stays 0..1 (1 for on, matching `core_getAllSol`). `GetSolenoid2VPMState` is untouched for the ext-board and GPIO callers.

Verified the emitted value in both modes:
- GoldenEye (`UseVPMModSol = 2`): flipper coils report 255, flippers work.
- Twister (no `UseVPMModSol`): flipper coils report 1, flippers work.

vbousquet:
Thanks!

gitfool:
@vbousquet thanks for your patience. I really appreciate it!

vbousquet:
I really appreciate your contribution too!

PR #648, opened 2026-09-05
wpc/sttng: add entry for LA-7 ROM

tomlogic:
Scorbit came across an LA-7 ROM in the wild, so here's the entry for PinMAME.  I've [uploaded to VPUniverse ](https://vpuniverse.com/files/file/31661-star-trek-the-next-generation-la-7/) and submitted a copy to VP Forums.

Is there interest in the "Americas" versions of LX WPC ROMs?  Scorbit needs to know about them because we're supporting users with those ROMs and need to recognize them, but if LX is better than LA, then maybe there isn't a point in adding them to PinMAME?


toxieainc:
Thanks! In general we're interested in all sets, no matter what. So there is no 'interesting', we're going for 100% preservation.  :)

PR #650, opened 2026-09-05
VPinMAME: scale WSHDlg layout by the dialog DPI

kara2010:
The script dialog API (`VPinMAME.WSHDlg`, used by core.vbs `cvpmDips` for the table option
menus many tables open on first start) positions its controls in pixels laid out for 96 dpi:
`AddCtrl` coordinates go straight into `CreateWindow`. The dialog font comes from the
DLGTEMPLATE (`DS_SETFONT`, Helv 8) and *is* scaled by Windows.

In a DPI unaware host Windows scaled the whole dialog, so this never showed. Visual Pinball
10.8.1 declares per-monitor DPI awareness, and from 150 % on the controls overlap their
own text. At 300 % the TAF options menu is unreadable:

before

This scales positions, sizes, the button gap and the dialog size/position by the dialog's
DPI (`GetDpiForWindow` via `GetProcAddress`, `GetDeviceCaps` fallback for older Windows).
DPI unaware hosts report 96 dpi and are unchanged.

after

Verified with a per-monitor-aware test process creating the same dialog as TAF, before and
after (the screenshots above), and in play with The Addams Family on Windows 11 at 300 %.

Disclosure: built with AI assistance under my direction; tested by hand.

### DIFF
--- a/src/win32com/WSHDlg.cpp
+++ b/src/win32com/WSHDlg.cpp
@@ -198,6 +198,32 @@ STDMETHODIMP CWSHDlg::Show(LONG_PTR hParentWnd, VARIANT *RetVal)
 
 #define CTRLID_START 1000
 
+// The script API positions controls in pixels laid out for 96 dpi. In a DPI aware host process (Visual Pinball
+// 10.8.1 and later declare per-monitor awareness), Windows no longer scales the dialog for us, but it still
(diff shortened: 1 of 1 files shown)


toxieainc:
Great, thanks, can cross that off my own todo list then  ;)

PR #651, opened 2026-09-05
sleic: get Bike Race V4.1 (bikerac3) working, and an opt-in ball model for Bike Race

gerwout:
Makes the Bike Race V4.1 set (`bikerac3`) work — it was registered
`GAME_NOT_WORKING` because it rendered garbage artwork — and adds an opt-in ball
model so all three Bike Race sets can be started standalone without holding
matrix keys.

## Bike Race V4.1 renders correctly now

The garbage was a **bad ROM 06 dump**, not a driver problem. `bikerac3`'s ROM 06,
CRC `ad48a30a`, is a defective read (confirmed with Joerg): its first `0x1104` bytes (the sprite/character
table) are the parent chip's own data with several `0x200`-byte pages duplicated,
so the 80188 reads a malformed record at the first sprite and runs off into the
blit loop.

Evidence, and how it was confirmed:

- ROM 04's far pointers into the sprite table are **byte-identical** to the 1992
  set's, and they point at addresses the good chip has records at and the bad dump
  does not.
- A logging read handler shows both sets fetching the same 120 bytes and then
  splitting on one: at flat `0x2001E` the parent reads `08 00 01 00 08 00` (a valid
  8×8 record) and the bad set reads `18 06 26 26 3C 3C`, a width of 1560 px.
- No Bike Race revision has ever changed a graphics ROM — ROM 05 is byte-identical
  across `bikerace`, `bikerac2` and V4.1, and it read perfectly in the same dump.
- **A re-dump of ROM 06 from the same V4.1 machine came back CRC `9db436d4`,
  byte-identical to the parent's `bkcpu06`,** passing every check the bad read
  failed.

So V4.1 is a genuine three-chip clone (`03`/`04`/`07` over the 1992 set), and this
PR inherits `bkcpu06` for ROM 06 and clears `GAME_NOT_WORKING`. The driver requires
CRC `9db436d4` and rejects the bad `ad48a30a` image (`bkcpu06.bin WRONG CHECKSUMS`).

Full write-up, disassembly and captures:
https://github.com/gerwout/sleic-iomoon/tree/main/asm/bikerace-2026-09

## Opt-in ball model for Bike Race

Standalone, Bike Race sits on "FALTA n BOLAS" until its trough optos are closed,
which otherwise means holding two matrix keys for the whole session. This mirrors
Io Moon's existing opt-in trough model: the `Balls` setting is the switch,
**default 0 = off** so a frontend/table still owns the optos and nothing is
fabricated; any non-zero value turns it on, and a new "Ball out of trough" cabinet
key lifts it. One `0xE0` opto mask satisfies all three sets' ball-status handlers.

Also in here (comment-only): the SLEIC3 memory-map comment had MCS1/MCS2 labelled
with each other's chip, and the COL4 trough optos are now named from the firmware's
own contact table (C6 "Salida Bolas", C7 "Bola en Espera", C8 "Bola Fuera").

## Testing (Linux, SDL3 build)

- All seven SLEIC sets pass `validitychecks` under the `MAME_DEBUG` build.
- `bikerace`, `bikerac2`, `bikerac3` boot, attract, take coins and play; V4.1's
  DMD output matches the parent frame-for-frame.
- `bikerac3` loads with no ROM warnings and starts without `-skip_gamewarnings`.

## Note: a pre-existing sound regression on Linux (not addressed here)
 
@toxieainc Unrelated to this PR, but worth flagging: on Linux, **Bike Race and Sleic Pin-Ball
FM audio plays slow/degraded on current builds**, and it was fine when the
Pin-Ball/Bike Race sets were first added (#627). Some change merged since then
introduced it. Io Moon (added later) is unaffected, so it splits by driver family
(SLEIC1/SLEIC3 affected, SLEIC2 not). The YM3812 register streams are byte-identical
between the old and new builds, so the emulation is generating the same data — the
regression is downstream (timing/output), not in the game drivers. This PR neither
causes nor fixes it; noting it here for visibility. I only confirmed this in Linux (Kali), so I did
not confirm if this is Linux only.

### DIFF
--- a/src/sound/3812intf.c
+++ b/src/sound/3812intf.c
@@ -107,6 +107,18 @@ static void YM3812Write_ymfm(int n, int a, int v)
 {
 	if (a & 1) /* data port */
 	{
+		/* Render the stream up to this write's CPU time before applying it, exactly as the
+		 * legacy fmopl path always did (fmopl.c OPLWrite calls its UpdateHandler, i.e.
+		 * stream_update, before every data write) and as the YM2151 ymfm wrapper does
+		 * (2151intf.c YM2151_data_port_0_w -> YM2151UpdateRequest).  Without it every
+		 * write of a video frame lands on the chip at one instant.  ymfm detects a
+		 * key-on/key-off EDGE only per generated sample (ymfm_fm.ipp clock_keystate),
+		 * so a key-off immediately followed by a key-on inside the same frame -- how
+		 * sequencers re-strike a repeated note and retrigger 0xBD drums -- collapses to
+		 * key 1->1: no attack, the old note just sustains.  fmopl re-attacks at write
+		 * time and never showed this.  Address writes only set the register latch and
+		 * are left unflushed, matching fmopl. */
+		stream_update(stream_3812[n], 0);
 		vgm_write(vgm_idx_3812[n], 0x00, lastreg_3812[n], v);
 		ymfm_opl_write(chip_3812[n], 1, v);
 	}
--- a/src/wpc/sleic.c
+++ b/src/wpc/sleic.c
@@ -1376,8 +1376,8 @@ MEMORY_END
  *   MCS0 0x00000-0x1FFFF : work RAM (boot stack 012F:0203; the boot copies its IVT
  *                          image from CS:00C4 to physical 0 before STI, so the
  *                          driver must NOT overwrite the IVT)
- *   MCS1 0x20000-0x3FFFF : graphics ROM bkcpu05
- *   MCS2 0x40000-0x5FFFF : graphics ROM bkcpu06 (read via ES=0x5000)
+ *   MCS1 0x20000-0x3FFFF : graphics ROM bkcpu06 (sprite table in its first 0x1104)
+ *   MCS2 0x40000-0x5FFFF : graphics ROM bkcpu05 (read via ES=0x5000)
  *   MCS3 0x60000-0x7FFFF : DMD / video frame buffer RAM (panel staging at 0x60410)
  * PACS=0xA03C -> peripheral block at 0xA0000, so sleic_periph_r/w are used.
  * UMCS -> bkcpu04 code at 0xE0000-0xFFFFF (reset EA F000:0000). */
--- a/src/wpc/sleic.c
+++ b/src/wpc/sleic.c
@@ -650,6 +650,7 @@ static void sleic_debug_switches(int troughCol, UINT8 troughDefault) {
   static int frame = 0;
   const char *e, *col, *bit;
   frame++;
+  if (getenv("SLEIC_TRACE_SND") && (frame % 60) == 0) fprintf(stderr, "[frame %d]\n", frame);
 
   if ((e = getenv("SLEIC_INJECT_CAB")))
     coreGlobals.swMatrix[9] |= (UINT8)strtol(e, NULL, 0);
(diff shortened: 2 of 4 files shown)


toxieainc:
@gerwout The only things that look like it could be causing something: fbe8a822f81951a8263fe6694231e68bb399c03c, 31c4adfdd91a005405ae28973bad55990775c685, and your own IO Moon changes.

gerwout:
@toxieainc  **Follow-up on the FM regression I noted above — it's the fmopl → ymfm YM3812 core switch, not the SLEIC driver afaik.**

I traced the "slow/degraded FM" and it is **not** in the SLEIC driver code (my Io Moon
changes included). The YM3812 emulation core changed from **fmopl** to **ymfm** as a
side effect of the build-system migration:

- The old `makefile.unixsdl` build — what #627 was validated against — compiles YM3812
  via `SOUNDS += YM3812@` → `HAS_YM3812=1` → the **fmopl** core (its `3812intf.o`
  references `YM3812UpdateOne`).
- The `cmake/sdl3pinmame` build (added in d4485612) defines `HAS_YM3812_YMFM=1` → the
  **ymfm** core. Since #647 removed the SDL 1.2 / makefile builds, everyone now gets ymfm.

Same register writes, different OPL2 implementation.

**Isolation:** the OKI MSM6376 goes through the same mixer / resampler / output path and
shares every build-level difference (SDL3, cmake, flags) — and it's fine. Only the FM
changed. So it isn't SDL3, the resampler, or compile flags; it's specifically the YM3812
core. I also confirmed the SLEIC YM3812 write path, the 4 MHz clock and the 55555→44100
stream rate are identical between the two builds, and that the YM3812 register stream the
driver emits is byte-identical.

This isn't SLEIC-specific — it affects **every YM3812 game** on the cmake build. It just
surfaces here because Bike Race / Pin-Ball can be A/B'd against the fmopl build (#627) and
against a real machine, whereas IO Moon only ever existed under ymfm.

I am digging into this to see if I can resolve this issue.

toxieainc:
I tested all YM3812 games when i changed it, and there it worked fine AFAIR. Obviously IO Moon was not there yet at that point back then.  ;)
In general, all platforms used in practice were already using the YMFM cores, it was only kept for pinmame to support ancient mingw's. And so far we didn't hear any complains from endusers.
Could still be, of course, that the YM3812 core has a bug, but its also what MAME changed to, and in general seen as a good tradeoff between perf and accuracy at the moment.

gerwout:
@toxieainc 
This one turned out to be harder to run down than it looked, and it ended up being two separate problems that happened to line up. The fix is the two commits just pushed to this branch.

**1. A broader YM3812 issue in the ymfm wrapper (`src/sound/3812intf.c`, e217bcf0).** This is the one behind the "missing FM sounds" that started after the fmopl -> ymfm switch (49fa7c29). It is not a bug in the ymfm core itself, so no disagreement with the choice to move to it: it is PinMAME's wrapper around it. `YM3812Write_ymfm` forwarded register writes to the chip without first rendering the audio stream up to the write's CPU time, so every write of a video frame landed on the chip at one instant. The old fmopl path never had that problem, because `fmopl.c OPLWrite` calls its `UpdateHandler` (`stream_update`) before every data write and re-attacks a voice at write time. ymfm instead only detects a key-on/key-off *edge* per generated sample, so a key-off immediately followed by a key-on inside the same frame, which is how a sequencer re-strikes a repeated note or retriggers the 0xBD rhythm bits, collapsed to "key still on": no new attack, the old note just sustained. The YM2151 ymfm wrapper already flushes before writes (`YM2151_data_port_0_w`); the YM3812 one did not.

**Does this affect other games?** Yes, by design: everything that goes through `3812intf.c`, which in PinMAME is Alvin G, JVH,  and the Sleic machines. The direction is towards what those games sounded like on fmopl (which they ran on until 49fa7c29) and towards the real chip. A game whose sound code never re-strikes a note within a single video frame is essentially unchanged; one that does gets those attacks back. Io Moon is a concrete example: it was losing roughly a third of its melodic re-attacks the same way, so "sounds fine on ymfm" was being judged against the collapsed rendering. I'd suggest a quick re-listen on the Alvin G / JVH / Mephisto titles; I'd expect nothing worse and some things better, but that is your call to make, and I'm happy to split this commit out into its own PR if you'd rather review it separately from the Sleic work.

**2. Bike Race driver issues (`src/wpc/sleic.c`, f4a92f87).

- The firmware runs the MSM6376 as two concurrent channels. The channel is not a bit of the phrase byte (every phrase write is masked with `AND AL,07F` first, and bit 7 of the phrase latch is unconnected on the board); it is PCS0 bit 3 (2CH on the IC32 control latch), sampled at the /ST rising edge. The driver collapsed both channels onto one core voice, so the second of every overlapping pair was refused by `adpcm.c` as a start on a still-playing voice. That is exactly how START issues the 4.7 s motor sample on channel 2 while the first sample is still running on channel 1, which is why that sound was never heard in emulation. Now both channels reach both core voices. 
- The YM3812's A0 line is a latched control bit (IC32 2Q, PCS0 bit 1), the same scheme Pin-Ball is modelled with, not a per-write hardware toggle. The old toggle happened to agree with the latched bit on every reachable write of this ROM, so this is fidelity rather than a fix, and the register stream is byte-identical before and after.

toxieainc:
Great, thanks!!

PR #654, opened 2026-09-06
unix/remote_debug: terminate headless runs and accept scripted input

gerwout:
-frames_to_run now ends a headless run: the display path stopped on the frame count but the headless path never consulted it, so -ftr ran forever. Both paths now spell the test the same way.

Adds cabinet-switch injection and a GI read-back to the remote debugger, so an unattended harness can drive dedicated buttons that have no matrix cell and read back general-illumination state. Drops input_port_get_force(), which had no callers.


toxieainc:
@whaslbeck Can you please review this?

whaslbeck:
Useful enhancements; looks good.

PR #657, opened 2026-09-06
i8051: fix per-source interrupt enables, stale vectors, and the edge-mode latch

gerwout:
Needs #PR656. This is the one needing scrutiny — it changes 15 of the 36 8051 games.

Three defects: each source's own IE enable bit is now required; a proposed vector that isn't dispatched is discarded (the old core could commit a stale V_IE1 while INT1 wasn't requesting and Timer 0 was — MAME keeps that vector in a function local for exactly this reason); and IE0/IE1 survive CLEAR_LINE in edge mode, which MCS-51 clears only in level mode. The core's own dispatch site already stated that rule, then destroyed the latch on release.

Evidence, 35-game sweep with -DI8051_SWEEP, NVRAM cleared, both sides re-baselined: 27 of 35 byte-identical, 0 of 35 audio digests changed — including every game that moved. On all 8 movers, dispatched_before + dropped = dispatched_after exactly. uboat65 had been discarding half its INT1s.

Blast radius measured two ways: 7 Alvin G DMD and uboat65 on IT1, 7 Spinball DMD on IT0; the other 20 are level-triggered and take a bit-identical path.

### DIFF
--- a/src/cpu/i8051/i8051.c
+++ b/src/cpu/i8051/i8051.c
@@ -1608,8 +1608,24 @@ INLINE UINT8 check_interrupts(void)
 #endif
 
 	//Skip the interrupt request if currently processing is lo priority, and the new request IS NOT HI PRIORITY!
+	//A proposal that is not dispatched must not survive the call.  Every entry to
+	//check_interrupts() re-proposes from the live flags, so discarding it here
+	//loses nothing; leaving it set makes a LATER call dispatch a vector whose flag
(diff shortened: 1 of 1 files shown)


toxieainc:
can you please rebase for clarity? thanks!

gerwout:
@toxieainc, done -> These i8051 changes should (apart from adding support for Mephisto and Sport 2000) also stabilize a lot of existing games.

PR #661, opened 2026-09-11
New driver: Recel System III (Rockwell PPS-4/2) — Fair Fight playable

gerwout:
Adds support for Recel System III, the Rockwell PPS-4/2 platform used by the
Spanish manufacturer Recel (Petaco S.A., Madrid) from 1978. Fair Fight is
playable start to finish; the remaining titles are not in this PR.

MAME has a `recel.cpp` skeleton with every set marked `NOT_WORKING`. This is a
from-scratch driver, but ROM set names match MAME's (hence `r_fairfght`).

### Hardware emulated

| Part | Role |
|---|---|
| Rockwell 11660 (PPS-4/2) | CPU, ~198.9 kHz |
| Rockwell 10788 GPKD | 32-position display controller |
| Rockwell 11696 PIO | 24 outputs: sound, coils, bonus BCD, indicators |
| A1761 / A1762 RRIOT | playfield lamps, NVRAM and printer control |
| HM6508 + CD4040 | 1024×1-bit serial CMOS, externally address-counted |
| Discrete sound | six PIO lines, summing rather than selecting |

The BIOS lives in the two RRIOT mask ROMs and is shared by every machine; each
game adds a small PROM read through bus interface chips that invert both the
address and the data bus, so it is decoded at init (`recel_decode_prom()`).

### What works

Self-check completes with the documented code sequence, then attract mode. All
three coin chutes, start, a full game to GAME OVER, scoring, bonus, playfield
lamps, the ball-in-play/game-over and match indicators, all four score
counters, credit and free-play/extra-ball displays, sound, and CMOS
adjustments that persist.

Two details worth calling out, both settled against a recording of a real
cabinet rather than inferred:

- **Score counters are six digits.** The GPKD multiplexes five; the 095-105
  unit's sixth 7448 position is wired to a permanent `0` (Recel scores in
  tens), so it is synthesised.
- **Lamp polarity.** The A1762 holding flip-flop reads 1 for OFF. Measuring
  both SPECIAL inserts across a gameplay recording shows them dark in 88-99%
  of frames; the opposite reading would light them for most of every game.

Adjustments are stored one nibble per RAM cell in the CMOS and restored at
power-up. A blank CMOS reads 0 throughout, which the ROM takes as "2 coins per
play" on chute 1 and "0 plays per coin" on chute 3 — a dead chute. That is
faithful to an unprogrammed board but not a useful shipped state, so a first
run (no NVRAM file) is seeded with 1 coin = 1 play and the three balls per
player the machine's instruction card specifies. A CMOS image that is
genuinely zero still restores as zero.


toxieainc:
Great!!!

PR #663, opened 2026-09-11
recel: add Mr. Evil (1978) and Mr. Doom (1979)

gerwout:
Both run on the same System III board as Fair Fight and share a playfield: Mr. Doom is a 1979 re-theme of Mr. Evil with the same 29-switch matrix and the same nine driver outputs, so they need no code beyond a ROM and a table entry. Verified against service docs 035-626 and 035-634/635: self-check, coining, a full game, the printed point values, and the drop-target bank rules (bank reset lights the 50,000/extra-ball award, second reset doubles bonus) all match.

The display layout is Fair Fight's -- four six-digit player counters plus the credit/ball/match digits -- because the 10788 GPKD wiring is a System III board feature, not a per-game one.


toxieainc:
Great, thanks!

Issue #662, opened 2026-09-11
Pinball Magic (Capcom) : flashing stobes / PF Back Flashers MX continously flashing


Dicentim974:
Hi,
Sorry I have this issue on the Pinball Magic table : my strobes and PF back Flashers are continously loop flashing.
I see you solved the issue here : https://github.com/vpinball/pinmame/issues/636
But I don't understand cause the bug is still here.
Please note that revert to official 3.6 release solves the issue.
Thanks for you help and patience

vbousquet:
Fixed in latest. Thanks for the report!

Issue #636, opened 2026-09-02
Capcom/SAM: solenoid legacy-fallback loop indexing bug


gitfool:
## Summary

The "revert to legacy 2-state integrator" fallback loop in `MACHINE_INIT(cc)` (Capcom) and `MACHINE_INIT(sam)` (Stern SAM) has an indexing bug: it iterates over every solenoid but always calls `core_set_pwm_output_type` with the constant `CORE_MODOUT_SOL0` (solenoid 1) instead of `CORE_MODOUT_SOL0 + i`, and it reads `physicOutputState[i]` instead of `physicOutputState[CORE_MODOUT_SOL0 + i]`. As a result the loop never reverts the intended per-game flasher outputs, and instead only ever rewrites solenoid 1.

Correcting the indexing is straightforward, but it exposes a second question about the loop's intended scope that I do not think an outside contributor should decide, hence this issue rather than a PR.

## Where

Both drivers have an identical copy of the loop.

`src/wpc/capcom.c`, `MACHINE_INIT(cc)` (around lines 758-762 at 6589fc57):

```c
// Defaults to 2 state legacy integrator for better backward compatibility
if ((options.usemodsol & (CORE_MODOUT_ENABLE_PHYSOUT_SOLENOIDS | CORE_MODOUT_ENABLE_MODSOL)) == 0)
   for (int i = 0; i < coreGlobals.nSolenoids; i++)
      if (coreGlobals.physicOutputState[i].type != CORE_MODOUT_SOL_2_STATE)
         core_set_pwm_output_type(CORE_MODOUT_SOL0, 1, CORE_MODOUT_LEGACY_SOL_2_STATE);
```

`src/wpc/sam.c`, `MACHINE_INIT(sam)` (around lines 1591-1595) has the same three lines.

## Why it is wrong

`CORE_MODOUT_SOL0` is `CORE_MODOUT_LAMP_MAX`, so it is not 0. Solenoid `i` lives at `physicOutputState[CORE_MODOUT_SOL0 + i]`, not `physicOutputState[i]`. So the loop has two defects:

1. The condition reads `physicOutputState[i]`, which indexes into the lamp region rather than the solenoids.
2. The body always passes `CORE_MODOUT_SOL0` with count 1, so regardless of `i` it only ever rewrites solenoid 1.

[...]

PR #670, opened 2026-09-13
capcom: keep pmv # 89 flasher bulbs on bulb integrator

gitfool:
## Problem

Pinball Magic (`pmv`, Capcom) shows its playfield flashers strobing on and off continuously ([#662](https://github.com/vpinball/pinmame/issues/662)). The reporter found that reverting to the 3.6 release restores the correct look.

The strobing is a side effect of the #636 backward-compatibility loop now that #636 works correctly. In `MACHINE_INIT(cc)`, the default-config fallback loop downgrades every output that is not `CORE_MODOUT_SOL_2_STATE` to `CORE_MODOUT_LEGACY_SOL_2_STATE`. The `pmv` flashers are typed `CORE_MODOUT_BULB_89_20V_DC_WPC`, so the loop rewrites them off the bulb integrator onto the legacy 2-state model. The legacy model renders each ROM pulse as a hard on/off blink instead of the bulb's thermal glow, which is the continuous strobing users see. Before #636 the loop was doubly broken and never touched these outputs, so 3.6 happens to look correct.

This regression was called out while #636 was being reviewed: [#636 (comment)](https://github.com/vpinball/pinmame/issues/636#issuecomment-5515355377) flagged that once the read index is corrected, the loop reverts every non-2-state output, which is the Option 1 vs Option 2 point from the original report.

## Fix

Two edits in `MACHINE_INIT(cc)`:

1. Exclude bulb-typed outputs from the legacy downgrade, so the per-game `#89` flasher bulbs keep the bulb integrator while the other outputs still revert to legacy.
2. Widen the `pmv` flasher range from 11 to 12 outputs. The driver typed sols 21-31, but the Pinball Magic operator manual (p.76, "Location of Solenoids, Motors, & Flashers") lists refs 21-32 as bulb outputs (part number `**` = `#89` bayonet base / 906 wedge base), with ref 32 being the SKILL/CHAM flasher. Sol 32 was left on the 2-state path and kept strobing even after the first edit. This second point is a pre-existing driver off-by-one, distinct from the #636/#662 regression, but both are needed for the table to look right.

This keeps the existing behavior for everything else, including the `SOL_CUSTOM` GameOn output (sol 51), which still reverts to `CORE_MODOUT_LEGACY_SOL_2_STATE` as it did before. Preserving that output as the driver types it was the Option 2 variant raised in [#636 (comment)](https://github.com/vpinball/pinmame/issues/636#issuecomment-5515355377); this change does not touch it, so that case is still open if you want it handled.

## Testing

Built libpinmame from this branch and ran `pmv112` in VPX on macOS arm64 (Metal).

Instrumented `MACHINE_INIT(cc)` to dump each solenoid's final output type in the default config (`usemodsol=0x80`, `FORCE_ON` only):

- Before: sols 21-31 = `301` (bulb), sol 32 = `2` (2-state), sol 51 = `50` (legacy, the `SOL_CUSTOM` GameOn).
- After: sols 21-32 = `301` (bulb), sol 51 = `50`, all real coils = `2`.

Visually confirmed in-game: the flashers no longer strobe continuously. Each flasher stays dark at rest and runs its ROM-driven pattern (a brief pulse or a few seconds of flashing during an award), which matches real hardware.

## Should SAM get the same fix?

`sam.c` (`MACHINE_INIT(sam)`) has the identical fallback loop, also noted in [#636 (comment)](https://github.com/vpinball/pinmame/issues/636#issuecomment-5515355377). Now that this table exposed the issue, should the same bulb-exclusion change be applied to the SAM driver? The count fix here is `pmv`-specific, but the loop change would carry over the same way for any SAM game that types per-game bulb flashers. Happy to fold it into this PR or handle it separately.


vbousquet:
Thanks !

If you identify that SAM needs some kind of identical fix, that would be nice to

Issue #664, opened 2026-09-12
No Good Gofer (NGG) wheel is not working correctly


apophis79:
**Describe the bug**
The visible mechanical wheel in the NGG playfield does not align with the position detected by the ROM. I believe the issue is related to a bug in PinMame's ngg.c file where the wheel is configured such that one revolution is 960 raw units, but the hardware switch model assumes it is 640 raw units. The wheel seems to have been configured correctly at one point, but it was changed [HERE](https://github.com/vpinball/pinmame/commit/5e62da5f9009b6041e96fd500408ce884888f7f2). 

**To Reproduce**
1. Download the table here: https://vpuniverse.com/files/file/28028-no-good-gofers-williams-1997
2. Add `debug.print WheelNewPos` within the `SRPRoutine` sub routine to monitor the wheel position reported to VPX. The attached video shows where I placed that print statement.
3. Run the table and open the debugger window to monitor the position.
4. Go into the ROM test menu and start the wheel test. The attached video shows where this is. 
5. Run wheel tests. Compare final value of the WheelNewPos reported in the debugger window with the position reported on the DMD. They are not always equal.

Attached demo video:
https://youtu.be/dfBwflSXLkc

**Expected behavior**
WheelNewPos should always equal the wheel position seen by the ROM. 

**Versions & Settings**
 - VPinMame v3.7.0-477-6589fc5
 - Windows version 11

vbousquet:
Fixed in latest. Thanks for the clear report !

PR #671, opened 2026-09-14
No Good Gofers: fix wheel position tracking

gitfool:
Fixes the No Good Gofers wheel position mismatch reported in #664, where the
visible wheel award decal did not align with the position the ROM detected.

The wheel mech in ngg.c was configured with `steps = 96 * NGG_WHEELRES`, but the
opto switch table in the same struct and the driver's position readout are built
for a 64-state wheel (`64 * NGG_WHEELRES`). Commit 5e62da5 changed this field from
64 to 96 in 2016 without adjusting the opto table, so the reported wheel position
was scaled by 96/64 relative to the optos the ROM reads, and the two drifted apart
across a revolution.

This reverts the `steps` field to `64 * NGG_WHEELRES`. The wheel speed change from
the same commit (NGG_SPEED and MECH_ACC 32 to 48) is unrelated to the position
scale and is left in place.

Verified in the operator wheel test: before the fix, DMD position 0 reported raw
867 and position 10 reported raw 562. After the fix, sampled DMD positions 2, 3, 9,
13, and 15 each match the driver's reported logical position, and across a
full-revolution capture the reported logical position covered all 16 values over
raw range 0..639.

Thanks to @apophis79 for the clear repro steps in #664, which made this
straightforward to reproduce and confirm.


gitfool:
Unrelated transient build failure:
```
Error: fatal: unable to access 'https://github.com/vpinball/pinmame/': Could not resolve host: github.com
  Error: The process '/opt/homebrew/bin/git' failed with exit code 128
  ```

vbousquet:
Thanks !

PR #676, opened 2026-09-14
efo: select the ZSU power-up ROM bank

gerwout:
The two lines that select one of the four 32K banks at `8000-FFFF` are pulled high through RP2 (2 x 10K to +5 on the ZSU schematic), so the board comes up on bank 3.

Nothing selected a bank at reset, and because the write handler only calls `cpu_setbank` when the value differs from its cached copy, a first write of bank 0 was suppressed as well and the window kept whatever it happened to point at.

Select bank 3 in `init()`, and move the cached bank number into `sndlocals` so it is cleared with the rest of the board state rather than persisting in a function-static across a reset.

Bank mapping is unchanged and matches MAME: bank 0 = `sonido2`, 1 = `sonido3`, 2 = `sonido4`, 3 = the unpopulated fifth socket. This is consistent with the Phantom Ship sound ROM itself, whose sample table needs exactly three banks and whose bank byte is `0xFF` (`0xFF & 3 = 3`).

Tested on Phantom Ship: full three-ball game, sound plays throughout.


toxieainc:
other machines also come up with the door open, but one can close it, and that state is then also saved. isn't that the case here?

gerwout:
@toxieainc You're right on both counts, and I tested them.

The toggle state does persist. With a dedicated cfg directory: pressed the door key, inserted 2 coins → credits `02`, exited cleanly; restarted with the same cfg and no door press, 2 more coins → credits `04`. It's saved in `phntmshp.cfg`, so it's a one-time action, not a per-session one.

And door-open-at-power-up is the established convention — `capcom.h`, `gts3.h` and `wpc.h` all declare "Coin Door" the same way.

So my description overstated the problem. "Unplayable out of the box" was wrong; it's a single keypress that then sticks. Apologies for the noise.

For what it's worth on the underlying behaviour, since I'd already set the tests up: on WPC an open door doesn't gate coins at all — T2 left at the default open state takes a coin and shows CREDITS 1 / PUSH START. On Playmatic series IV/V it does: coins are ignored and start goes to bookkeeping instead. That matches the machine's own manual, where bookkeeping is entered by opening the door and pressing the game button. So PinMAME is modelling the real behaviour correctly and the default is fine as it stands.

I've dropped that commit. The PR is now just the ZSU bank fix, which is independent: the two banking lines are pulled high through RP2 (2 x 10K to +5 on the schematic) so the board powers up on bank 3, but nothing selected a bank at reset, and a first write of bank 0 was suppressed by the `!=` check against the cached value.

PR #687, opened 2026-09-20
VPinMAME: add thread-safe main CPU byte read

kylehudson00:
## Summary

Adds a read-only `ReadMainCPUByte(address)` method to the VPinMAME COM interface, allowing clients such as VPX table scripts to read a byte from the main CPU memory space.

Memory reads are marshalled to the emulation thread and performed with CPU 0's memory context active using `cpunum_read_byte()`. This avoids relying on the memory context that happens to be active on the calling thread.

The implementation also:

- avoids queueing when already running on the emulation thread
- validates the requested address against the main CPU address space
- returns `-1` when the emulator is not running, the address is invalid, or the read cannot be completed within the timeout
- keeps successful return values in the `0..255` range

## Motivation

This method enables VPX table scripts to access ROM state that is not reliably exposed through switches, lamps or solenoids. This opens up new possibilities for table authors, allowing access to things like whether a game is currently active, player count, current ball, high scores and so on. My immediate need was to add forced single-player modes to ROM-based VPX tables, which can only be reliably implemented by referencing a ROM's internal `game_over` flag or similar, but there are many possible use cases.

Exposing a generic read-only main CPU memory API allows the table script to decide which ROM address and semantics are appropriate without adding game-specific logic to VPinMAME.

## Acknowledgement

The thread-safe memory access approach in this change was adapted from the LibPinMAME work in PR #643, in particular dispatching memory reads to the emulation thread, explicitly selecting the CPU context, and reading through `cpunum_read_byte()` rather than directly dereferencing memory.

This PR applies that approach to the traditional Windows VPinMAME COM interface.

## VPX Usage

`ReadMainCPUByte(address)` returns the byte at the specified address in the main CPU's memory space. Successful reads return a value from `0` to `255`; `-1` indicates that the value could not be read.

The address and interpretation of a value are ROM-specific. A useful resource for locating game-state variables is the Pinball Memory Maps project:

https://github.com/tomlogic/pinball-memory-maps

ROM list: https://github.com/tomlogic/pinball-memory-maps/blob/main/maplist.md

The project documents CPU memory locations for many PinMAME ROM sets, including game-state values such as `game_over`, `current_player`, `player_count`, `current_ball`, and `tilted`. The map should be checked for the particular ROM set being used, and values should be verified where appropriate.

Once the relevant address is known, a VPX script can read it directly:

```vb
state = Controller.ReadMainCPUByte(address)

If state <> -1 Then
    ' Interpret state according to the ROM's memory map.
End If
```

### Example: The Addams Family (TAF_L7)

For the Williams `TAF_L7` ROM, Pinball Memory Maps identifies address `$008D` as `game_over`.

This can be read from a VPX script as follows:

```vb
Dim gameOver
gameOver = Controller.ReadMainCPUByte(&H8D)

If gameOver = 1 Then
    ' Game over / attract mode
ElseIf gameOver = 0 Then
    ' Game is in progress
Else
    ' Read unavailable
End If
```

A VPX table that needs to prevent the Start button from adding additional players once a game has begun could use:

```vb
If keycode = StartGameKey Then
    If Controller.ReadMainCPUByte(&H8D) = 0 Then Exit Sub
End If
```

The address and value semantics in this example are specific to `TAF_L7`; other ROM sets should use the appropriate entry from their memory map.

### VBScript hexadecimal addresses

When using hexadecimal addresses in VPX VBScript, literals at or above `&H8000` should use the VBScript Long suffix (`&`).

For example:

```vb
Controller.ReadMainCPUByte(&H8000&)
Controller.ReadMainCPUByte(&HFFFF&)
```

Without the trailing `&`, VBScript can interpret these literals as negative signed values (for example, `&H8000` as `-32768` and `&HFFFF` as `-1`). `ReadMainCPUByte` intentionally rejects negative addresses.

This was verified during testing: decimal addresses `32768` and `65535`, and their suffixed hexadecimal equivalents `&H8000&` and `&HFFFF&`, read correctly.

## Testing

Built successfully as an x64 Release build with Visual Studio 2026.

Runtime memory reads and game-state use were tested with:

- Williams WPC: The Addams Family and Attack from Mars
- Data East/Sega Version 3b: Batman Forever
- Stern Whitestar: The Simpsons Pinball Party
- Gottlieb System 80: Haunted House

Testing included attract mode, active gameplay, Tilt, ball/player transitions, match/game-over states, repeated reads, fixed-ROM reads, and table restart.

The final rebased build was also installed and smoke-tested successfully in VPX.


jayadelson:
@kylehudson00 I guess we should decide which of the two approaches (#643 or #685 ) is best, since this one is following the #643 mechanism. I got the impression @toxieainc objected to that so I created #685, or someone (maybe @toxieainc or @vbousquet ) can settle which is preferred. Unfortunately the PRs have sat for a few weeks now without comment, so I don't know how to proceed. If there is something I'm missing or a thread I need to jump on let me know.

kylehudson00:
> @kylehudson00 I guess we should decide which of the two approaches (#643 or #685 ) is best, since this one is following the #643 mechanism. I got the impression @toxieainc objected to that so I created #685, or someone (maybe @toxieainc or @vbousquet ) can settle which is preferred. Unfortunately the PRs have sat for a few weeks now without comment, so I don't know how to proceed. If there is something I'm missing or a thread I need to jump on let me know.

Thanks. I based #687 on the #643 approach because an earlier version I tried using `memory_get_read_ptr()` gave inconsistent results in testing, probably due to the issue described in #642. I’m happy to adapt #687 to whichever approach the maintainers prefer. Using the #685 approach could allow this PR to be simplified quite a bit. One point where some general guidance would be appreciated — should this method retain `cpunum_read_byte()` semantics for handler-backed regions, or would raw memory access via the corrected `memory_get_read_ptr()` be considered sufficient?

kylehudson00:
Update: I’ve revised #687 locally on top of #685. The revised implementation uses `memory_get_read_ptr()` for directly backed memory, with the existing emulation-thread `cpunum_read_byte()` path retained as a fallback where no direct pointer is available. I’ve confirmed both paths in testing. Will update this PR with revised implementation once #685 is merged.

PR #689, opened 2026-09-21
Recel: comparator display strip, full 7448 decode, coin door key, and Torneo

gerwout:
Four fixes to the Recel System III driver, plus Torneo as two sets.

Gerrit Volkenborn kindly passed on the preliminary Recel driver he abandoned
in 2024, and asked why a counter that runs to four figures in his driver reads
backwards and two digits wide in this one. Chasing that produced most of what
follows.

### The representation areas were unreadable

With the coin door open, the start button steps through sixteen RAM
representation areas and the displays stop being score counters. An area's
nibbles then land wherever the lite box happens to wire each map column, so a
four-digit value in columns 3..0 arrives split across a score counter, two
one-digit windows and a block of LEDs — which is exactly what the cabinet
does, and useless to read.

Recel shipped an answer: the **094-631 Lite Box Comparator** (doc 035-637), a
bench unit that replaces the lite box and represents all 32 nibbles as
seven-segment digits, in map-column order F..0. `recel_disp` now carries that
strip as two extra rows, one layout entry per group using `CORE_SEGREV`.

It is drawn in `CORE_SEG7S`: at 198 kHz the PPS-4 executes only ~3300
instructions per frame, so per-frame character drawing dominates the frame
cost. Measured over a fixed span of ROM on an idle machine, 32 full-size
digits cost **13%** of emulator throughput against **4%** for the small ones.

### Group B columns 8 and 9 were being discarded

The comparator manual's INDICATORS figure names all 32 positions. B8/B9 are
not spare — they are *"No of advances on counters and the last figure
affected"*, a technical indication with no cabinet indicator — so they are
decoded digits now rather than forced blank. The figure also names the four
indicator nibbles: SELECTED, IN PLAY, ONE MILLION POINTS, HANDICAP.

### Nibbles above 9 blanked in release builds

`core_bcd2seg7a[]` carries its A–E entries only under `MAME_DEBUG`, so a
release build decoded every nibble above 9 to zero. Invisible during play,
where the ROM sends BCD — wrong in the representation areas, which put raw RAM
on the displays. Replaced with a local `recel_seg7[]` covering all sixteen
codes, re-read from the factory glyph table: C is `b,f,g` rather than `c,d,e`
(the drawing's `∪` sits in the upper half of the cell), and 6 has no top bar
and 9 no bottom bar — both confirmed against a frame of a running cabinet
showing `9` and `6`.

### Coin door as a maintained switch

Door and tilt are one contact (§4.4), and the door is maintained: the test and
adjustment mode is entered by opening it, which a momentary key cannot hold.
`KEYCODE_7` — PinMAME's usual coin door key — now toggles it, `KEYCODE_DEL`
stays momentary for the tilt, and Fault moves to `KEYCODE_HOME`. The toggle has
to be the later of the two entries on bit 4, or `inptport.c`'s default-value
pass undoes the latch.

### Torneo, as two sets

`r_torneo` is the image that circulates. It plays **one** ball: PROM offset
`0xF6` decodes to `t $80B` at `0x809`, whose target is the second byte of the
`tl $2D0` at `0x80A`. Landing there executes that `0xD0` as `tm ($0D0)`, a call
to the ball-counter increment at BIOS `0x117`, and the return lands on `0x80C`
— another `tm ($0D0)`. Two increments per pass through the `0x800` hook, which
runs at game start *and* on every drain, so the counter passes the limit tested
at `0x3EA` on the first drain. Mr. Evil carries the identical test at `0x803`
and jumps clear of its own `tl $2D0` onto a real instruction.

`r_torneoa` is that byte repaired, `0x74 → 0x72`, i.e. `t $80D`, landing past
both `tm` bytes on the `lb ($0C5)` that forces the mode of play.

| set | balls before GAME OVER |
|---|---|
| `r_torneo` | 1 |
| `r_torneoa` | **3** |
| `r_fairfght` (control) | 3 |

Both hashes are flagged `BAD_DUMP`: the circulating image is demonstrably
wrong, and the repair is a reconstruction rather than a read of a physical
PROM. `0x74 → 0x72` is a two-bit difference, not a single-bit flip, and only
one image exists to compare against — lisy.dev's `torneo_26AB.BIN` is
byte-identical to garzol's `to.c5`, so one bad burn or read propagated to both.
A fresh read of a physical Torneo PROM would settle it.

Happy to attach the two ROM sets (`r_torneo.zip`, `r_torneoa.zip`) — say the
word and I will add them here.


gerwout:
[recel-torneo-roms.zip](https://github.com/user-attachments/files/32462792/recel-torneo-roms.zip)
Attached the original (bad) and corrected ROM

